PR #43, opened 2014-12-24
memory usage optimizations for read and write operations

mtuncer:


cstore_writer.c
```diff
@@ -1033,3 +1003,21 @@ SyncAndCloseFile(FILE *file)
 						errmsg("could not close file: %m")));
 	}
 }
+
+
+/* CopyStringInfo creates a deep copy of given source string */
+StringInfo
+CopyStringInfo(StringInfo sourceString)
```
mtuncer: I wanted to use as little space as needed. appendStringInfo tries to make sure it has enough space by calling enlargeStringInfo, which allocates as twice as memory needed trying to save from additional allocations that could be requested later.  so this function is not to save from line of code. it is only meant to allocate right amount of memory.

cstore_writer.c
```diff
@@ -239,9 +326,10 @@ CStoreWriteRow(TableWriteState *writeState, Datum *columnValues, bool *columnNul
 			Oid columnCollation = attributeForm->attcollation;
 
 			blockData->existsArray[blockRowIndex] = true;
-			blockData->valueArray[blockRowIndex] = DatumCopy(columnValues[columnIndex],
-															 columnTypeByValue,
-															 columnTypeLength);
+
+			SerializeSingleDatum(blockData->uncompressedDataBuffer,
+								 columnValues[columnIndex], columnTypeByValue,
+								 columnTypeLength, attributeForm->attalign);
```
mtuncer: although buffer enlargements seemed to be done incrementally, it does allocate in chunks i.e. twice as the existing memory amount. It preallocates 1024 bytes in the beginning, and doubles the memory as needed during enlarge operation.

cstore_writer.c
pykello: Makes sense. Please edit the function comments to make our intention explicit.

cstore_reader.c
```diff
@@ -42,6 +42,7 @@ static StripeData * LoadFilteredStripeData(FILE *tableFile,
 										   List *whereClauseList);
 static void ReadStripeNextRow(StripeData *stripeData, List *projectedColumnList,
 							  uint64 blockIndex, uint64 blockRowIndex,
+							  DeserializedColumnBlockData **deserializedColumnBlockDataArray,
```
pykello: Exceeds the 90 character limit. Maybe shorter names can help.

cstore_reader.c
```diff
@@ -55,6 +56,10 @@ static StripeSkipList * LoadStripeSkipList(FILE *tableFile,
 										   StripeFooter *stripeFooter,
 										   uint32 columnCount,
 										   Form_pg_attribute *attributeFormArray);
+DeserializedColumnBlockData**
+CreateColumnBlockDataBuffers(uint32 columnCount, bool* columnMask, uint32 blockRowCount);
+void FreeColumnBlockDataBuffers(DeserializedColumnBlockData **deserializedColumnBlockDataArray,
+						   uint32 columnCount);
```
pykello: 1. First line exceeds the 90 character limit.
2. Second line not aligned with first line.

cstore_reader.c
```diff
@@ -66,8 +71,8 @@ static StripeSkipList * SelectedBlockSkipList(StripeSkipList *stripeSkipList,
 											  bool *selectedBlockMask);
 static uint32 StripeSkipListRowCount(StripeSkipList *stripeSkipList);
 static bool * ProjectedColumnMask(uint32 columnCount, List *projectedColumnList);
-static bool * DeserializeBoolArray(StringInfo boolArrayBuffer, uint32 boolArrayLength);
-static Datum * DeserializeDatumArray(StringInfo datumBuffer, bool *existsArray,
+static void DeserializeBoolArray(StringInfo boolArrayBuffer, bool *boolArray, uint32 boolArrayLength);
```
pykello: Exceeds the 90 character limit.

cstore_reader.c
```diff
@@ -66,8 +71,8 @@ static StripeSkipList * SelectedBlockSkipList(StripeSkipList *stripeSkipList,
...
+static void DeserializeDatumArray(StringInfo datumBuffer, bool *existsArray, Datum *datumArray,
 									 uint32 datumCount, bool datumTypeByValue,
 									 int datumTypeLength, char datumTypeAlign);
```
pykello: 1. First line exceeds the 90 character limit.
2. Second and third lines not aligned with the first line.

cstore_reader.c
```diff
@@ -114,6 +121,12 @@ CStoreBeginRead(const char *filename, TupleDesc tupleDescriptor,
 											  ALLOCSET_DEFAULT_MINSIZE,
 											  ALLOCSET_DEFAULT_INITSIZE,
 											  ALLOCSET_DEFAULT_MAXSIZE);
+	columnCount = tupleDescriptor->natts;
+	projectedColumnMask = ProjectedColumnMask(columnCount, projectedColumnList);
+
+	DeserializedColumnBlockData** deserializedColumnBlockDataArray =
```
pykello: Definition in middle of block. Causes compile warnings.

cstore_reader.c
```diff
@@ -890,26 +1046,22 @@ DeserializeBoolArray(StringInfo boolArrayBuffer, uint32 boolArrayLength)
...
+ * stores them in provided datumArray. If a value is marked as false in the exists array, the
  * function assumes that the datum isn't in the buffer, and simply skips it.
  */
-static Datum *
-DeserializeDatumArray(StringInfo datumBuffer, bool *existsArray, uint32 datumCount,
+static void
+DeserializeDatumArray(StringInfo datumBuffer, bool *existsArray, Datum *datumArray, uint32 datumCount,
 					  bool datumTypeByValue, int datumTypeLength,
 					  char datumTypeAlign)
 {
```
pykello: 1. By looking at the function signature, the names ```existsArray``` and ```datumArray``` seem to be of same kind, but ```existsArray``` is an input argument and ```datumArray``` is a modify argument. I think we should make this more explicit. For example, we can rename ```datumArray``` to ```resultDatumArray```.
2. Input arguments and output arguments are mixed. I think we should either move all output arguments to the beginning of list or end of list. Please search our codes to see what we have used in other places.
3. First line exceeds the 85 character limit.

cstore_reader.c
```diff
@@ -890,26 +1046,22 @@ DeserializeBoolArray(StringInfo boolArrayBuffer, uint32 boolArrayLength)
...
 	}
-
-	return boolArray;
 }
 
 
 /*
  * DeserializeDatumArray reads an array of datums from the given buffer and
- * returns the array. If a value is marked as false in the exists array, the
+ * stores them in provided datumArray. If a value is marked as false in the exists array, the
```
pykello: Exceeds 85 character.

cstore_reader.c
```diff
@@ -858,13 +1016,12 @@ ProjectedColumnMask(uint32 columnCount, List *projectedColumnList)
...
 /*
- * DeserializeBoolArray reads an array of bits from the given buffer and returns
- * it as a boolean array.
+ * DeserializeBoolArray reads an array of bits from the given buffer and stores
+ * it in provided bool array.
  */
-static bool *
-DeserializeBoolArray(StringInfo boolArrayBuffer, uint32 boolArrayLength)
+static void
+DeserializeBoolArray(StringInfo boolArrayBuffer, bool *boolArray, uint32 boolArrayLength)
```
pykello: Please see the comments on output argument naming and ordering in ```DeserializeDatumArray```.

cstore_reader.c
```diff
@@ -427,17 +589,13 @@ LoadColumnData(FILE *tableFile, ColumnBlockSkipNode *blockSkipNodeArray,
...
 												 blockSkipNode->valueLength);
-		StringInfo valueBuffer = DecompressBuffer(rawValueBuffer,
-												  blockSkipNode->valueCompressionType);
 
-		Datum *valueArray = DeserializeDatumArray(valueBuffer, existsArray,
-												  rowCount, typeByValue, typeLength,
-												  typeAlign);
-		blockDataArray[blockIndex]->valueArray = valueArray;
+		blockDataArray[blockIndex]->serializedValueBuffer = rawValueBuffer;
+		blockDataArray[blockIndex]->valueCompressionType = blockSkipNode->valueCompressionType;
```
pykello: Exceeds 90 characters. Maybe a new variable can help. Like we did for ```rowCount```.

cstore_writer.c
```diff
@@ -177,17 +185,85 @@ CStoreBeginWrite(const char *filename, CompressionType compressionType,
...
+
+		serializedValueBuffer = deserializedBlockData->uncompressedDataBuffer;
+
+		serializedBlockData->valueCompressionType = COMPRESSION_NONE;
+		serializedBlockData->rowCount = rowCount;
+
+		if (compressionType == COMPRESSION_NONE)
+		{
+			serializedBlockData->serializedValueBuffer = CopyStringInfo(serializedValueBuffer);
+			continue;
```
mtuncer: this code block used to be here way before. I strongly agree with the comment.

cstore_fdw.h
```diff
@@ -170,6 +170,15 @@ typedef struct StripeSkipList
 } StripeSkipList;
 
 
+typedef struct BlockDataBuffers
```
pykello: 1. I think in other places we have used the term _buffer_ for raw data, and not for deserialized data. So I prefer renaming ```BlockDataBuffers``` to ```ColumnBlockData```, and renaming ```StripeData```, ```ColumnData```, and ```ColumnBlockData``` to ```StripeDataBuffers```, ```ColumnDataBuffers```, and ```ColumnBlockDataBuffers```.

2. We put a comment before each struct.

cstore_fdw.h
```diff
@@ -259,6 +271,8 @@ typedef struct TableWriteState
 	StripeData *stripeData;
 	StripeSkipList *stripeSkipList;
 	uint32 stripeMaxRowCount;
+	BlockDataBuffers **blockDataBuffersArray;
```
pykello: If we decide to change the name of ```BlockDataBuffers```, we should also change the field name here.

cstore_fdw.h
```diff
@@ -259,6 +271,8 @@ typedef struct TableWriteState
...
+	StringInfo decompressionBuffer;
```
pykello: Can we put a comment before this field to explain what is the reason for having ```decompressionBuffer```?

cstore_reader.c
```diff
@@ -114,6 +120,12 @@ CStoreBeginRead(const char *filename, TupleDesc tupleDescriptor,
 											  ALLOCSET_DEFAULT_MINSIZE,
 											  ALLOCSET_DEFAULT_INITSIZE,
 											  ALLOCSET_DEFAULT_MAXSIZE);
+	columnCount = tupleDescriptor->natts;
+	projectedColumnMask = ProjectedColumnMask(columnCount, projectedColumnList);
+
+	blockDataBuffersArray = CreateBlockDataBuffersArray(columnCount,
```
pykello: If we change ```BlockDataBuffers``` to ```ColumnBlockData```, we can also rename this function to ```CreateEmptyBlockDataArray``` or something similar.

cstore_reader.c
```diff
@@ -200,6 +214,85 @@ CStoreReadFooter(StringInfo tableFooterFilename)
 }
 
 
+/* DeserializeBlockData  deserializes requested data block for all columns. It
+ * uncompresses serialized data if necessary, sets deserialized buffer index at
+ * read state. The function also deallocates previous data buffer used to keep temporary
+ * uncompressed data buffer.
+ */
+static void
+DeserializeBlockData(TupleDesc tupleDescriptor, StripeData *stripeData,
```
pykello: 1. Forward declaration missing,
2. Defined before being used.

cstore_reader.c
```diff
@@ -200,6 +214,85 @@ CStoreReadFooter(StringInfo tableFooterFilename)
...
+		ColumnData *columnData = stripeData->columnDataArray[columnIndex];
+
+		if (columnData != NULL)
+		{
+			ColumnBlockData *blockData = columnData->blockDataArray[blockIndex];
+			Form_pg_attribute *attributeFormArray = tupleDescriptor->attrs;
+			bool typeByValue = attributeFormArray[columnIndex]->attbyval;
+			int typeLength = attributeFormArray[columnIndex]->attlen;
+			char typeAlign = attributeFormArray[columnIndex]->attalign;
+			BlockDataBuffers * blockDataBuffers = blockDataBuffersArray[columnIndex];
```
pykello: Extra space after ```*```.

cstore_reader.c
```diff
@@ -200,6 +214,85 @@ CStoreReadFooter(StringInfo tableFooterFilename)
...
+}
+
+
+/*
+ * ResetUncompressedDataBuffers iterates over deserialized column block data
+ * buffers and sets uncompressedDataBuffer field to NULL. This field is
+ * allocated in stripe memory context and becomes invalid once memory context
+ * is reset.
+ */
+static void
```
pykello: 1. Function comments doesn't match the function implementation,
2. Forward declaration missing,
3. Defined before being used.

cstore_reader.c
```diff
@@ -200,6 +214,85 @@ CStoreReadFooter(StringInfo tableFooterFilename)
 }
 
 
+/* DeserializeBlockData  deserializes requested data block for all columns. It
```
pykello: Extra space after ```DeserializeBlockData```.

cstore_reader.c
```diff
@@ -200,6 +214,85 @@ CStoreReadFooter(StringInfo tableFooterFilename)
...
+								 blockData->rowCount);
+			DeserializeDatumArray(valueBuffer, blockDataBuffers->existsArray,
+								  blockData->rowCount, typeByValue, typeLength,
+								  typeAlign, blockDataBuffers->valueArray);
+
+			/* store newly allocated uncompressed data buffer to be freed later */
+			if (valueBuffer != blockData->valueBuffer)
+			{
+				blockDataBuffers->serializedValueBuffer = valueBuffer;
+			}
```
pykello: 1. I think using ```if (blockData->valueCompressionType != COMPRESSION_NONE)``` might be a bit more explicit and more matching with the comment?
2. Why are we saving only the uncompressed data to be freed later? Can't we do this also for the non-compressed case?
3. If we can also do this for the compressed case, then we can initialize ```serializedValueBuffer``` with ```makeStringInfo()```, and remove the ```if (blockDataBuffers->serializedValueBuffer != NULL)``` for pfreeing it. This also removes the need for ```ResetUncompressedDataBuffers()```.

cstore_reader.c
```diff
@@ -200,6 +214,85 @@ CStoreReadFooter(StringInfo tableFooterFilename)
...
+	{
+		ColumnData *columnData = stripeData->columnDataArray[columnIndex];
+
+		if (columnData != NULL)
+		{
+			ColumnBlockData *blockData = columnData->blockDataArray[blockIndex];
+			Form_pg_attribute *attributeFormArray = tupleDescriptor->attrs;
+			bool typeByValue = attributeFormArray[columnIndex]->attbyval;
+			int typeLength = attributeFormArray[columnIndex]->attlen;
+			char typeAlign = attributeFormArray[columnIndex]->attalign;
```
pykello: It seems that we don't need ```attributeFormArray``` for anything other than getting ```typeByValue```, ```typeLength```, and ```typeAlign```. So I think we can also something like this which is a bit simpler:

```C
Form_pg_attribute attributeForm = tupleDescriptor->attrs[columnIndex];
bool typeByValue = attributeForm.attbyval;
int typeLength = attributeForm.attlen;
char typeAlign = attributeForm.attalign;
```

Since ```attributeForm.attbyval```, etc. is short enough we could also have remove the three variables (```typeByValue```, etc.) we have used.

cstore_reader.c
```diff
@@ -248,6 +341,9 @@ CStoreReadNextRow(TableReadState *readState, Datum *columnValues, bool *columnNu
 		{
 			readState->stripeData = stripeData;
 			readState->stripeReadRowCount = 0;
+			readState->deserializedBlockIndex = -1;
+			ResetUncompressedDataBuffers(readState->blockDataBuffersArray,
+										stripeData->columnCount);
```
pykello: Not aligned with the previous line.

cstore_reader.c
```diff
@@ -276,16 +382,76 @@ CStoreReadNextRow(TableReadState *readState, Datum *columnValues, bool *columnNu
...
+void
+FreeColumnBlockDataBuffers(BlockDataBuffers **blockDataBuffersArray,
+						   uint32 columnCount)
+{
+	uint32 columnIndex = 0;
+
+	for (columnIndex = 0; columnIndex < columnCount; columnIndex++)
+	{
+		BlockDataBuffers *blockDataBuffers = blockDataBuffersArray[columnIndex];
+
```
pykello: In cases when we check for a variable right after assignment we usually don't use new-lines between them.

cstore_fdw.h
```diff
@@ -170,6 +170,15 @@ typedef struct StripeSkipList
 } StripeSkipList;
 
 
+typedef struct BlockDataBuffers
+{
+	bool *existsArray;
+	Datum *valueArray;
+	StringInfo serializedValueBuffer;
```
pykello: 1. I think ```serializedValueBuffer``` is here because ```valueArray``` can have pointers to its buffer. So, I think the fact that this is result of serialization or not isn't important. So, can we rename this to ```valueBuffer```?

2. Can we put a comment here and explain why are we putting this here? It is not clear and can lead to confusion. (Our style for member field comments is: newline, then the comment, then the member field.)

cstore_fdw.h
```diff
@@ -177,8 +186,10 @@ typedef struct StripeSkipList
  */
 typedef struct ColumnBlockData
 {
-	bool *existsArray;
-	Datum *valueArray;
+	StringInfo existBuffer;
+	StringInfo valueBuffer;
```
pykello: One other argument for renaming ```ColumnBlockData``` to ```ColumnBlockDataBuffers``` is that we have used the member names ```existBuffer``` and ```valueBuffer``` inside it, so it is actually containing buffers.

pykello: ```existBuffer``` -> ```existsBuffer```?

cstore_fdw.h
```diff
@@ -177,8 +186,10 @@ typedef struct StripeSkipList
...
+	CompressionType valueCompressionType;
+	uint32 rowCount;
```
pykello: We can infer this from ```stripeData->rowCount```, ```blockRowCount```, and ```currentBlockIndex```, and we have only used this in only function, and this is the same value among all columns, so can we remove this from here to make it a bit simpler?

cstore_reader.c
```diff
@@ -114,6 +120,12 @@ CStoreBeginRead(const char *filename, TupleDesc tupleDescriptor,
 											  ALLOCSET_DEFAULT_MINSIZE,
 											  ALLOCSET_DEFAULT_INITSIZE,
 											  ALLOCSET_DEFAULT_MAXSIZE);
+	columnCount = tupleDescriptor->natts;
+	projectedColumnMask = ProjectedColumnMask(columnCount, projectedColumnList);
+
```
pykello: ```columnCount``` and ```projectedColumnMask``` are related to ```blockDataBuffersArray``` but not ```stripeReadContext```. So, can we move this new line to the line before ```columnCount```?

cstore_reader.c
pykello: Very minor. The only field from ```tupleDescriptor``` that we are using is ```atts```. Can we pass only that?

cstore_reader.c
```diff
@@ -200,6 +214,85 @@ CStoreReadFooter(StringInfo tableFooterFilename)
 }
 
 
+/* DeserializeBlockData  deserializes requested data block for all columns. It
+ * uncompresses serialized data if necessary, sets deserialized buffer index at
+ * read state. The function also deallocates previous data buffer used to keep temporary
+ * uncompressed data buffer.
```
pykello: Comment is not in sync with what function does. We are not passing read state to this function.

cstore_reader.c
```diff
@@ -200,6 +214,85 @@ CStoreReadFooter(StringInfo tableFooterFilename)
...
+ */
+static void
+DeserializeBlockData(TupleDesc tupleDescriptor, StripeData *stripeData,
+					 MemoryContext stripeReadContext, uint64 blockIndex,
+					 BlockDataBuffers **blockDataBuffersArray)
```
pykello: It is not clear from the signature what is the function's input and what is the output. Maybe we can do something like this:

```C
static void
DeserializeBlockData(StripeDataBuffers *stripeBuffers, uint32 blockIndex,
					 TupleDesc tupleDescriptor,
					 ColumnBlockData **outputBlockDataArray);
```

This passes what's going to be deserialized as the first argument, then the helper variables as the 2nd and 3rd arguments, and then finally the output argument with a distinctive name. Please make sure to use the same ordering and naming style that you have used for other similar functions. (e.g. should we use ```outputBlockDataArray``` or ```resultBlockDataArray```? etc.)

cstore_reader.c
```diff
@@ -200,6 +214,85 @@ CStoreReadFooter(StringInfo tableFooterFilename)
...
 
 
+/* DeserializeBlockData  deserializes requested data block for all columns. It
+ * uncompresses serialized data if necessary, sets deserialized buffer index at
+ * read state. The function also deallocates previous data buffer used to keep temporary
+ * uncompressed data buffer.
+ */
+static void
+DeserializeBlockData(TupleDesc tupleDescriptor, StripeData *stripeData,
+					 MemoryContext stripeReadContext, uint64 blockIndex,
```
pykello: ```stripeReadContext``` is not very related to deserialization and I think we can move it out of this function and do the context switches in the caller.

cstore_reader.c
```diff
@@ -200,6 +214,85 @@ CStoreReadFooter(StringInfo tableFooterFilename)
...
+					 BlockDataBuffers **blockDataBuffersArray)
+{
+	int columnIndex = 0;
+	MemoryContext oldContext = MemoryContextSwitchTo(stripeReadContext);
+
+	for (columnIndex=0; columnIndex < stripeData->columnCount; columnIndex++)
+	{
+		ColumnData *columnData = stripeData->columnDataArray[columnIndex];
+
+		if (columnData != NULL)
```
pykello: When we do ```X x = ...``` and then ```if (x ...)```, we usually avoid empty line between them.

pykello: We usually use ```continue``` for the exceptional cases that can be handled in the beginning of block. I think this is also OK according to Code Complete 2. One advantage of this reducing the indentation levels.

Since the indentation levels are not much in this case, keeping as it is also perfectly fine.

cstore_reader.c
```diff
@@ -200,6 +214,85 @@ CStoreReadFooter(StringInfo tableFooterFilename)
...
+ */
+static void
+DeserializeBlockData(TupleDesc tupleDescriptor, StripeData *stripeData,
+					 MemoryContext stripeReadContext, uint64 blockIndex,
+					 BlockDataBuffers **blockDataBuffersArray)
+{
+	int columnIndex = 0;
+	MemoryContext oldContext = MemoryContextSwitchTo(stripeReadContext);
+
+	for (columnIndex=0; columnIndex < stripeData->columnCount; columnIndex++)
```
pykello: ```columnIndex=0``` -> ```columnIndex = 0```

cstore_reader.c
```diff
@@ -890,26 +1049,22 @@ DeserializeBoolArray(StringInfo boolArrayBuffer, uint32 boolArrayLength)
...
+ * DeserializeDatumArray reads an array of datums from the given buffer and stores
+ * them in provided datumArray. If a value is marked as false in the exists array,
+ * the function assumes that the datum isn't in the buffer, and simply skips it.
  */
-static Datum *
+static void
 DeserializeDatumArray(StringInfo datumBuffer, bool *existsArray, uint32 datumCount,
 					  bool datumTypeByValue, int datumTypeLength,
-					  char datumTypeAlign)
+					  char datumTypeAlign, Datum *datumArray)
```
pykello: By looking at the function signature, the names ```existsArray``` and ```datumArray``` seem to be of same kind, but ```existsArray``` is an input argument and ```datumArray``` is a modify argument. I think we should make this more explicit. For example, we can rename ```datumArray``` to ```resultDatumArray```.

cstore_reader.c
```diff
@@ -858,13 +1018,13 @@ ProjectedColumnMask(uint32 columnCount, List *projectedColumnList)
...
- * DeserializeBoolArray reads an array of bits from the given buffer and returns
- * it as a boolean array.
+ * DeserializeBoolArray reads an array of bits from the given buffer and stores
+ * it in provided bool array.
  */
-static bool *
-DeserializeBoolArray(StringInfo boolArrayBuffer, uint32 boolArrayLength)
+static void
+DeserializeBoolArray(StringInfo boolArrayBuffer, bool *boolArray,
+					 uint32 boolArrayLength)
```
pykello: Ordering and naming of output variables should be consistent in the whole code.

cstore_writer.c
pykello: Can I explain this a bit more?

What I suggested here just moves out some lines out of the ```SerializeDatum``` so it's responsibility becomes less. The enlarge operation will be executed with the same arguments and the same number of calls as before.

cstore_writer.c
```diff
@@ -810,57 +723,123 @@ SerializeBoolArray(bool *boolArray, uint32 boolArrayLength)
...
-			Assert(!datumTypeByValue);
-			memcpy(currentDatumDataPointer, DatumGetPointer(datum), datumLength);
-		}
 
-		datumBuffer->len += datumLengthAligned;
-	}
+		/* store (compressed) value buffer */
+		serializedBlockData->valueCompressionType = actualCompressionType;
+		serializedBlockData->valueBuffer =
+				CopyStringInfo(serializedValueBuffer);
```
pykello: Can be merged with the previous line.

cstore_writer.c
```diff
@@ -810,57 +723,123 @@ SerializeBoolArray(bool *boolArray, uint32 boolArrayLength)
...
+	 * check and compress value buffers, if a value buffer is not compressable
+	 * then keep it as uncompressed, store compression information.
+	 */
+	for (columnIndex = 0; columnIndex < columnCount; columnIndex++)
+	{
+		uint64 maximumLength = 0;
+		bool compressable = false;
+		ColumnData *columnData = stripeData->columnDataArray[columnIndex];
+		ColumnBlockData *serializedBlockData =
+				columnData->blockDataArray[blockIndex];
```
pykello: Can be merged with the previous line.

cstore_fdw.h
mtuncer: ColumnBlockDataBuffers is getting too long. we will have problems fitting into single line when we have statement like
ColumnBlockDataBuffers *columnBlockDataBuffers = columnDataBuffers->blockDataBufferArray[blockIndex];

this is 101 character long if I haven't miscounted. 

could we shorten them a bit ?

cstore_fdw.h
mtuncer: ok

cstore_writer.c
mtuncer: ok

cstore_fdw.h
mtuncer: per our discussion on shortening names
StripeData becomes StripeBuffers 
ColumnData becomes ColumnBuffers
ColumnBlockData becomes ColumnBlockBuffers

cstore_fdw.h
mtuncer: done. it became `ColumnBlockData **blockDataArray`

cstore_reader.c

cstore_reader.c
mtuncer: added forward declaration and moved function definition downwards. 

Since Form_pg_attribute also used in other functions' interfaces I changed here too.

cstore_reader.c

cstore_reader.c
mtuncer: I think this is the same as previous commend regarding DeserializeBlockData. No extra action is taken for this one

cstore_reader.c
mtuncer: 1. agreed. we should also make it more explicit by calling DecompressBuffer() only if compression type is something other than COMPRESSION_NONE
2.  Since DecompressBuffer allocates memory for each decompression operation within stripe read memory context, buffers used for decompressing may pile up until memory context reset at next stripe read. This way we could reclaim memory during block changes. Non-compressed case is a bit different. Memory is already read into serialized data buffers (ColumnBlockBuffers->valueBuffer), Deserialized version (ColumnBlockData->valueArray) may contain pointers to serialized version. since it was already in memory, there is no additional memory is allocated and there is no need to free it right away.

3.  I did not understand the comment here. Could you explain a bit more ?

cstore_reader.c
mtuncer: ok

cstore_reader.c
mtuncer: looked well locally, realigned again

cstore_reader.c
mtuncer: that space was for variable declaration. Could you re-evaluate this comment ? thanks

cstore_fdw.h
mtuncer: done

cstore_reader.c
pykello: Here the comment says "... and sets uncompressedDataBuffer field to NULL" while we are actually setting ```serializedValueBuffer``` to ```NULL``` and not ```uncompressedDataBuffer```. Other details of the comment should be double checked, just to make sure.

cstore_reader.c
pykello: In CitusDB's source code, when we do ```X x = ...```, we usually don't separate it from ```if (x ...)``` with newline. i.e. we do something like:

```C
BlockDataBuffers *blockDataBuffers = blockDataBuffersArray[columnIndex];
if (blockDataBuffers != NULL)
{
 ...
}
```

instead of

```C
BlockDataBuffers *blockDataBuffers = blockDataBuffersArray[columnIndex];

if (blockDataBuffers != NULL)
{
 ...
}
```

I don't the reason why we're doing this always. I don't know if there is much value in being consistent in these little details.

cstore_reader.c
pykello: For 2, (1) if it doesn't hurt the correctness of program, (2) and if freeing the buffer in uncompressed case makes the code a bit simpler (less if's, less functions, etc.), then we can do something like this:

```C
/* free previous block's data buffers */
pfree(blockDataBuffers->serializedValueBuffer->data);
pfree(blockDataBuffers->serializedValueBuffer);

/* decompress and deserialize current block's data */
valueBuffer = DecompressBuffer(blockData->valueBuffer,
							   blockData->valueCompressionType);
if (blockData->valueCompressionType != COMPRESSION_NONE)
{
	/* we don't need the original value buffer anymore */
	pfree(blockData->valueBuffer->data);
	pfree(blockData->valueBuffer);
}

DeserializeBoolArray(blockData->existBuffer,
					 blockDataBuffers->existsArray,
					 blockData->rowCount);
DeserializeDatumArray(valueBuffer, blockDataBuffers->existsArray,
					  blockData->rowCount, typeByValue, typeLength,
					  typeAlign, blockDataBuffers->valueArray);

/* save current block's data buffer to be freed when we're done with the block */
blockDataBuffers->serializedValueBuffer = valueBuffer;
```

We can even remove the ```if (blockData->valueCompressionType != COMPRESSION_NONE)``` block without hurting the correctness, but it may confuse the reader that we're freeing the original buffer for non-compressed, and not for compressed.

If we model the code as above, then we need to make some other changes (which is what I wanted to say in 3):

1. In ```CreateBlockDataBuffersArray()```, instead of initializing ```serializedValueBuffer``` to ```NULL```, we should initialize it to ```makeStringInfo()```. This is because in the above code we are unconditionally freeing ```serializedValueBuffer```.
2. We probably can remove ```ResetUncompressedDataBuffers```.

I am not sure if what I said plays well with memory contexts. Here I am trying to make the code a bit simpler. If you think this doesn't make the code simpler, please think if we can make the code simpler in any other way or not.

Thanks.

cstore_reader.c
mtuncer: ok

cstore_reader.c

cstore_reader.c
mtuncer: Thanks for the clarification
I made most of the changes with 1 exception.
I had to keep ResetUncompressedDataBuffers() function because after processing the first stripe,  although memory is freed with context reset, blockDataBuffer->serializedValueBuffer still points to some memory location. CreateBlockDataBuffersArray is called only once when beginning reading from cstore table and not called again. 

Thinking about if we can move this deserialized block data structure into stripe read context too.

cstore_reader.c
mtuncer: OK, I will do that way.

but this requires more changes in other places.

cstore_fdw.h
mtuncer: done

cstore_fdw.h
mtuncer: done, row count is removed from here and accessed via alternate means

cstore_reader.c

cstore_reader.c

cstore_reader.c
mtuncer: we did not in/out type in other functions. Doing it here would create a rule for future. Are you sure about this ? May be we should discuss it with Ozgun and team.

Would it be ok if we just put that information on function comments ?

cstore_reader.c

cstore_reader.c
mtuncer: after previous refactoring nesting is reduced by 1 level

cstore_reader.c

cstore_reader.c
mtuncer: We need a coding rule here.
I usually do following
- if integer specifically contains the length of the buffer, it comes after the buffer declaration
- if integer has more meanings like having number of items in certain collection to be put inside a provided buffer, it may come before buffer declaration.

I made quick check on citusDB codebase but did not find enough examples to reach a decision.

cstore_writer.c
mtuncer: done

Issue #44, opened 2015-01-06
text column error when using "select ... into cstore_fdw_table from ...."


nickinchina:
My source table has text columns which contain a big chunk of text and after i move them to cstore_fdw tables, i can't read it back, i got the following error:

ERROR: missing chunk number 0 for toast value 17007 in pg_toast_16445
SQL status: XX000

mtuncer:
Hello
Thanks for reporting the issue. 
would you mind sharing more data that would help us investigate the problem ?
- table schema
- insert/select statements
- sample data if possible

thanks

pykello:
I could generate a related error, but not the exact error:

```SQL
CREATE TABLE text_data AS
SELECT string_agg(random()::text, '')
FROM generate_series(1, 1000) a, generate_series(1, 10) b
GROUP BY b;

CREATE FOREIGN TABLE cstore_text_data(A text) SERVER cstore_server;

INSERT INTO cstore_text_data SELECT * FROM text_data;

SELECT char_length(a) FROM cstore_text_data;

DROP TABLE text_data;

SELECT char_length(a) FROM cstore_text_data;

ERROR:  could not open relation with OID 16522
```

We are looking into this and will resolve it.

mtuncer:
this issue is fixed and merged into development branch. it will be available in master branch at upcoming release (1.2)

pykello:
Merged into master. Closing.

PR #45, opened 2015-01-07
fixes issue #44 datum is detoasted for variable length attributes

mtuncer:
matching regression test is added


pykello:
This looks good. I will do some testing tomorrow, then we can check this in.

PR #61, opened 2015-06-30
Row count estimation (into develop)

mtuncer:


pykello:
This looks good to me. I'll merge it after some testing.

PR #62, opened 2015-07-01
Fixed the deadlock issue for concurrent inserts.

pykello:
Previously we were using Exclusive locks to protect concurrent INSERTs from each other. But since PostgreSQL acquires RowExclusive locks automatically before executing cstore's insert logic, and Exclusive & RowExclusive locks conflict, this resulted in deadlock.

This pull request fixes this by using ShareUpdateExclusive locks, which is self-conflicting, but doesn't conflict with RowExclusive locks.


mtuncer:
looks good to me.

tested with concurrent insert, select, count(), copy from operations thru pgbench, also run concurrent drop foreign table. works fine.

PR #64, opened 2015-07-07
alter table command is now supported on cstore_fdw

mtuncer:


cstore_reader.c
```diff
@@ -675,6 +679,30 @@ LoadStripeSkipList(FILE *tableFile, StripeMetadata *stripeMetadata,
...
+
+		/* create empty ColumnBlockSkipNode for missing columns*/
+		columnSkipList = palloc0(stripeBlockCount * sizeof(ColumnBlockSkipNode));
+
+		for (blockIndex = 0; blockIndex < stripeBlockCount; blockIndex++)
+		{
+			columnSkipList->rowCount = 0;
+			columnSkipList->hasMinMax = false;
+			columnSkipList->minimumValue = (Datum)NULL;
+			columnSkipList->maximumValue = (Datum)NULL;
```
pykello: I think we usually use "0" instead of "(Datum)NULL" for datums.

cstore_reader.c
```diff
@@ -1101,22 +1129,26 @@ DeserializeDatumArray(StringInfo datumBuffer, bool *existsArray, uint32 datumCou
  * DeserializeBlockData deserializes requested data block for all columns and
  * stores in blockDataArray. It uncompresses serialized data if necessary. The
  * function also deallocates data buffers used for previous block, and compressed
- * data buffers for the current block which will not be needed again.
+ * data buffers for the current block which will not be needed again. If a column
+ * data is not present serialized buffer, than default value (or null) is used
```
pykello: "than" -> "then"?

cstore_reader.c
```diff
@@ -1101,22 +1129,26 @@ DeserializeDatumArray(StringInfo datumBuffer, bool *existsArray, uint32 datumCou
...
-					 ColumnBlockData **blockDataArray)
+					 ColumnBlockData **blockDataArray, TupleDesc tupleDescriptor)
 {
 	int columnIndex = 0;
 	for (columnIndex = 0; columnIndex < stripeBuffers->columnCount; columnIndex++)
 	{
+		ColumnBlockData *blockData = blockDataArray[columnIndex];
+		Form_pg_attribute attributeForm = attributeFormArray[columnIndex];
 		ColumnBuffers *columnBuffers = stripeBuffers->columnBuffersArray[columnIndex];
+		bool columnAdded = (columnBuffers == NULL) && (blockData != NULL);
```
pykello: When I wrote similar boolean assignment earlier, Ozgun suggested to use the following form instead:

```C
bool columnAdded = false;
if ((columnBuffers == NULL) && (blockData != NULL))
{
	columnAdded = true;
}

File: cstore_reader.c [hunk 4]
```diff
@@ -1144,7 +1176,62 @@ DeserializeBlockData(StripeBuffers *stripeBuffers, uint64 blockIndex,
 			/* store current block's data buffer to be freed at next block read */
 			blockData->valueBuffer = valueBuffer;
 		}
+		else if (columnAdded)
+		{
+			/* populate exists and value buffers for missing column */
+			int rowIndex = 0;
+			bool useDefaultValue = attributeForm->atthasdef;
+			Datum defaultValue = (Datum) NULL;
```
pykello: We usually use "0" for datums instead of "(Datum) NULL".

cstore_reader.c
```diff
@@ -1144,7 +1176,62 @@ DeserializeBlockData(StripeBuffers *stripeBuffers, uint64 blockIndex,
...
+			}
+		}
+	}
+}
+
+
+/*
+ * DefaultValueForColumn returns default value for given column.
+ * Only const values are supported. The function returns NULL for all other
+ * default value types.
```
pykello: I'm not sure, but should we error out for non-const default value types of added columns so users don't get incorrect results?

cstore_reader.c
```diff
@@ -1144,7 +1176,62 @@ DeserializeBlockData(StripeBuffers *stripeBuffers, uint64 blockIndex,
...
+ */
+static bool
+DefaultValueForColumn(TupleConstr *tupleConstraints, int columnIndex, Datum *defaultValue)
+{
+	int defaultIndex = 0;
+	int defaultCount = tupleConstraints->num_defval;
+	bool found = false;
+
+	*defaultValue = (Datum) NULL;
```
pykello: "0" instead of "(Datum) NULL".

pykello:
I added some comments. Everything else looks good. We should also make the coverage checks pass. We can look into it later.

cstore_reader.c

cstore_reader.c

cstore_reader.c
mtuncer: ok. I want to check code complete's say on this.

cstore_reader.c

cstore_reader.c
mtuncer: I am more into checking this during alter command as we discussed. Per your suggestion I will check immutable functions too

cstore_reader.c

cstore_reader.c
```diff
@@ -1144,10 +1183,108 @@ DeserializeBlockData(StripeBuffers *stripeBuffers, uint64 blockIndex,
...
+														columnIndex, columnName,
+														&defaultValue);
+			}
+
+			for (rowIndex = 0; rowIndex < rowCount; rowIndex++)
+			{
+				blockData->existsArray[rowIndex] = useDefaultValue;
+				/* it is safe to assign datum value since this array content
+				 * is never deallocated.
+				 */
```
pykello: Comment doesn't match our coding styles. The text of comment should start from 2nd line and the it should start with a capital letter.

```C
/*
 * It is safe to assign datum value since this array content
 * is never deallocated.
 */
```

pykello:
I did a quick review. I'll do a thorougher review next week before checking in, but I think there aren't any major issues left.

cstore_reader.c
mtuncer: ok. fixed and resubmitted

PR #67, opened 2015-08-12
added truncate table support

mtuncer:
it only supports truncating a foreign table. truncate options are not supported, silently ignored.

### DIFF
--- a/cstore_fdw.c
+++ b/cstore_fdw.c
@@ -66,7 +66,8 @@ static uint64 CopyIntoCStoreTable(const CopyStmt *copyStatement,
 								  const char *queryString);
 static uint64 CopyOutCStoreTable(CopyStmt* copyStatement, const char* queryString);
 static List * DroppedCStoreFilenameList(DropStmt *dropStatement);
-static void TruncateCStoreTable(TruncateStmt *truncateStatement);
+static List * FindCStoreTables(List *tableList);
+static void TruncateCStoreTables(List *cstoreTableList);
 static void DeleteCStoreTableFiles(char *filename);
 static void InitializeCStoreTableFile(Oid relationId);
 static bool CStoreTable(Oid relationId);
@@ -237,9 +238,18 @@ CStoreProcessUtility(Node *parseTree, const char *queryString,
 	}
 	else if (nodeTag(parseTree) == T_TruncateStmt)
 	{
-		TruncateCStoreTable((TruncateStmt*) parseTree);
-		CallPreviousProcessUtility(parseTree, queryString, context,
-								   paramListInfo, destReceiver, completionTag);
+		TruncateStmt *truncateStatement = (TruncateStmt *) parseTree;
+		List *allTableList = truncateStatement->relations;
+		List *cstoreTableList = FindCStoreTables(allTableList);
+		List *otherTableList = list_difference(allTableList, cstoreTableList);
+		if (list_length(otherTableList) > 0)
+		{
+			truncateStatement->relations = otherTableList;
+			CallPreviousProcessUtility(parseTree, queryString, context, paramListInfo,
+									   destReceiver, completionTag);
+		}
+
+		TruncateCStoreTables(cstoreTableList);
 	}
 	/* handle other utility statements */
 	else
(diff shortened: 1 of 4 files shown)


cstore_fdw.c
```diff
@@ -251,6 +233,13 @@ CStoreProcessUtility(Node *parseTree, const char *queryString,
 
 			DeleteCStoreTableFiles(fileName);
 		}
+
+	}
+	else if (nodeTag(parseTree) == T_TruncateStmt)
+	{
+		TruncateCStoreTable((TruncateStmt*) parseTree);
+		CallPreviousProcessUtility(parseTree, queryString, context,
+								   paramListInfo, destReceiver, completionTag);
```
pykello: 1. TruncateCStoreTable -> TruncateCStoreTables?
2. I was confused here for a while that why the standard utility is not throwing error here, and that was because I couldn't tell from the name "TruncateCStoreTable" that it removes cstore tables from parseTree->relations. Maybe we should break TruncateCStoreTable into 2 or more functions to make it more explicit.
3. What happens when we do ```TRUNCATE cstore_table, regular_table, file_fdw_table```? I guess we truncate the cstore table, then standard truncate errors out and doesn't truncate any of the other tables?
4. ```(TruncateStmt*) parseTree``` -> ```(TruncateStmt *) parseTree```? (added a space character)

Given above, what about something like this?
```C
	else if (nodeTag(parseTree) == T_TruncateStmt)
	{
		TruncateStmt *truncateStatement = (TruncateStmt *) parseTree;
		List *cstoreTableList = FindCStoreTables(truncateStatement->relations);

		/*
		 * Truncate non-cstore tables. If an error happens here, we won't continue
		 * with cstore tables.
		 */
		truncateStatement->relations = list_difference(truncateStatement->relations,
													   cstoreTableList);
		CallPreviousProcessUtility(parseTree, queryString, context,
								   paramListInfo, destReceiver, completionTag);

		/* truncate cstore tables */
		TruncateCStoreTables(cstoreTableList);
	}
```

Tuxie:
Thank you for implementing this! We really would like to have support for TRUNCATE, because without it, we can't refresh a table without breaking grants, views and functions referring to it. I hope to see this in a release build soon!

cstore_fdw.c
```diff
@@ -532,6 +531,45 @@ DroppedCStoreFilenameList(DropStmt *dropStatement)
...
+}
+
+/* TruncateCStoreTable truncates given cstore tables */
+static void
+TruncateCStoreTables(List *cstoreTableList)
+{
+	ListCell *relationCell = NULL;
+	foreach(relationCell, cstoreTableList)
+	{
+		RangeVar *rangeVar = (RangeVar *) lfirst(relationCell);
```
onderkalaci: In postgres documentation (and source) it says that `truncate` [acquires](http://www.postgresql.org/docs/9.5/static/sql-truncate.html) `ACCESS EXCLUSIVE lock `. I think we should definitely acquire AccessExcluiveLock as we acquire `ExclusiveLock` [here](https://github.com/citusdata/cstore_fdw/blob/feature-9.5_compatibility/cstore_fdw.c#L199). 
Or, maybe we can use `RangeVarGetRelid ` with `AccessExclusiveLock`, which seems equivalent, but I am not sure.

(Note: Postgres acquires the lock as [here](https://github.com/postgres/postgres/blob/ee943004466418595363d567f18c053bae407792/src/backend/commands/tablecmds.c#L1022))

cstore_fdw.c
```diff
@@ -251,6 +234,22 @@ CStoreProcessUtility(Node *parseTree, const char *queryString,
 
 			DeleteCStoreTableFiles(fileName);
 		}
+
+	}
+	else if (nodeTag(parseTree) == T_TruncateStmt)
+	{
+		TruncateStmt *truncateStatement = (TruncateStmt *) parseTree;
+		List *allTableList = truncateStatement->relations;
+		List *cstoreTableList = FindCStoreTables(allTableList);
```
onderkalaci: Maybe `allTablesList`? and similarly `cstoreTablesList`?

cstore_fdw.c
```diff
@@ -566,6 +604,30 @@ DeleteCStoreTableFiles(char *filename)
 
 
 /*
+ * InitializeCStoreTableFile creates data and footer file for a cstore table.
+ * The function assumes data and footer files do not exist, therefore
```
onderkalaci: Is there a way to add `Assert` statements for the function's assumptions?

cstore_fdw.c
onderkalaci: btw, I was able to crash backend with concurrent truncate & copy commands with the lack of `AccessExclusiveLock`.

cstore_fdw.c
```diff
@@ -532,6 +531,45 @@ DroppedCStoreFilenameList(DropStmt *dropStatement)
...
+
+/* TruncateCStoreTable truncates given cstore tables */
+static void
+TruncateCStoreTables(List *cstoreTableList)
+{
+	ListCell *relationCell = NULL;
+	foreach(relationCell, cstoreTableList)
+	{
+		RangeVar *rangeVar = (RangeVar *) lfirst(relationCell);
+		Oid relationId = RangeVarGetRelid(rangeVar, AccessShareLock, true);
```
onderkalaci: One minor note, cstore does nothing on the following `truncate` [parameters](http://www.postgresql.org/docs/9.5/static/sql-truncate.html): `[ RESTART IDENTITY | CONTINUE IDENTITY ] [ CASCADE | RESTRICT ]`.
Do you think should we notice the users if those parameters exist (ie: if any parameter exists)?

I know `CASCADE` and `RESTRICT` cannot even be applicable since they rely on table constraints, where `cstore` does not allow at all.

cstore_fdw.c
```diff
@@ -251,6 +236,22 @@ CStoreProcessUtility(Node *parseTree, const char *queryString,
...
 		}
+
+	}
+	else if (nodeTag(parseTree) == T_TruncateStmt)
+	{
+		TruncateStmt *truncateStatement = (TruncateStmt *) parseTree;
+		List *allTableList = truncateStatement->relations;
+		List *cstoreTableList = FindCStoreTables(allTableList);
+		List *otherTableList = list_difference(allTableList, cstoreTableList);
+		if (list_length(otherTableList) > 0)
```
pykello: ```otherTableList != NIL```?

cstore_fdw.c
```diff
@@ -532,6 +533,47 @@ DroppedCStoreFilenameList(DropStmt *dropStatement)
...
+		if (CStoreTable(relationId))
+		{
+			cstoreTableList = lappend(cstoreTableList, rangeVar);
+		}
+	}
+
+	return cstoreTableList;
+}
+
+/* TruncateCStoreTable truncates given cstore tables */
```
pykello: Two empty lines between function definitions.

cstore_fdw.c
mtuncer: done, however I am seeing deadlocks in the repetitive copy/truncate commands. Same also happens in regular postgresql tables.

cstore_fdw.c

cstore_fdw.c
mtuncer: There is but I do not think it is meaningful to add many file functions here.

cstore_fdw.c
mtuncer: really :) will do that.

cstore_fdw.c
mtuncer: done

onderkalaci:
@mtuncer The important points of my review feedback is applied. I re-tested it, and it's a :ship: 

(btw, I'll open an issue on concurrent COPY & DROP & TRUNCATE scenario as we talked in person)

PR #78, opened 2016-01-12
Feature 9.5 compatibility

mtuncer:
Few API changes in PG95 broke cstore_fdw. With this change cstore_fdw builds and works with both PG95 and pre-PG95 versions.

### DIFF
--- a/cstore_compression.c
+++ b/cstore_compression.c
@@ -1,0 +1,165 @@
+/*-------------------------------------------------------------------------
+ *
+ * cstore_compression.c
+ *
+ * This file contains compression/decompression functions definitions
+ * used in cstore_fdw.
+ *
+ * Copyright (c) 2016, Citus Data, Inc.
+ *
+ * $Id$
+ *
+ *-------------------------------------------------------------------------
+ */
+#include "postgres.h"
+#include "cstore_fdw.h"
+
+#if PG_VERSION_NUM >= 90500
+#include "common/pg_lzcompress.h"
+#else
+#include "utils/pg_lzcompress.h"
+#endif
+
+
+
+
+#if PG_VERSION_NUM >= 90500
+/*
+ *	The information at the start of the compressed data. This decription is taken
+ *	from pg_lzcompress in pre-9.5 version of PostgreSQL.
+ */
+typedef struct CStoreCompressHeader
+{
+	int32		vl_len_;		/* varlena header (do not touch directly!) */
+	int32		rawsize;
+} CStoreCompressHeader;
+
+/*
+ * Utilities for manipulation of header information for compressed data
+ */
+
+#define CSTORE_COMPRESS_HDRSZ		((int32) sizeof(CStoreCompressHeader))
+#define CSTORE_COMPRESS_RAWSIZE(ptr) (((CStoreCompressHeader *) (ptr))->rawsize)
+#define CSTORE_COMPRESS_RAWDATA(ptr) (((char *) (ptr)) + CSTORE_COMPRESS_HDRSZ)
+#define CSTORE_COMPRESS_SET_RAWSIZE(ptr, len) (((CStoreCompressHeader *) (ptr))->rawsize = (len))
+
+#else
+
+#define CSTORE_COMPRESS_HDRSZ		(0)
+#define CSTORE_COMPRESS_RAWSIZE(ptr) (PGLZ_RAW_SIZE((PGLZ_Header *) buffer->data))
+#define CSTORE_COMPRESS_RAWDATA(ptr) (((PGLZ_Header *) (ptr)))
+#define CSTORE_COMPRESS_SET_RAWSIZE(ptr, len) (((CStoreCompressHeader *) (ptr))->rawsize = (len))
+
+#endif
+
+
+
+/*
+ * CompressBuffer compresses the given buffer with the given compression type
+ * outputBuffer enlarged to contain compressed data. The function returns true
+ * if compression is done, returns false if compression is not done.
+ * outputBuffer is valid only if the function returns true.
+ */
+bool
+CompressBuffer(StringInfo inputBuffer, StringInfo outputBuffer,
+			   CompressionType compressionType)
+{
+	uint64 maximumLength = PGLZ_MAX_OUTPUT(inputBuffer->len) + CSTORE_COMPRESS_HDRSZ;
+	bool compressionResult = false;
+
+	if (compressionType != COMPRESSION_PG_LZ)
+	{
+		return false;
+	}
+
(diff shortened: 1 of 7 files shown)


jasonmp85:
Hey @mtuncer: I just noticed the Coveralls settings for `cstore_fdw` were not set correctly. I have `pg_shard` set to only fail a build if coverage drops by more than 0.5% during a single change or if it drops beneath 90% overall. Dropping by 0.02% is basically noise, so this shouldn't have been marked as a bad build. I fixed the settings and am re-triggering the build.

jasonmp85:
Closed to re-trigger a Travis build to test settings.

cstore_compression.c
```diff
@@ -0,0 +1,165 @@
...
+
+
+
+
+#if PG_VERSION_NUM >= 90500
+/*
+ *	The information at the start of the compressed data. This decription is taken
+ *	from pg_lzcompress in pre-9.5 version of PostgreSQL.
+ */
+typedef struct CStoreCompressHeader
```
onderkalaci: Can't we move this struct definition, the following `#define`s and `CompressBuffer` & `DecompressBuffer` declartions to `cstore_compression.h` ? I think I'd be more compact for compression related stuff.

Also, we generally use `define` macros in the headers and having the declarations in `cstrore_fdw.h` doesn't look the best approach.

cstore_compression.c
```diff
@@ -0,0 +1,165 @@
+/*-------------------------------------------------------------------------
+ *
+ * cstore_compression.c
+ *
+ * This file contains compression/decompression functions definitions
+ * used in cstore_fdw.
+ *
+ * Copyright (c) 2016, Citus Data, Inc.
```
onderkalaci: Since we get `CStoreCompressHeader ` definition from older postgres, should we add `Portions Copyright`? I really don't know this, but asking if it is necessary or not.

cstore_compression.c
```diff
@@ -0,0 +1,165 @@
...
+		return false;
+	}
+
+	resetStringInfo(outputBuffer);
+	enlargeStringInfo(outputBuffer, maximumLength);
+
+#if PG_VERSION_NUM >= 90500
+	int32 compressedByteCount = 0;
+
+	compressedByteCount = pglz_compress((const char *) inputBuffer->data,
```
onderkalaci: Ok, I read&understand the API changes. But, I have a question for our implementation. What happens when `pglz_compress ` returns `0`. Is there a possibility that `outputBuffer->data` is `NULL` etc so that cstore crashes?

I wasn't able to test it, do you know how to do it? If so, can you test that?

cstore_compression.c
```diff
@@ -0,0 +1,165 @@
...
+	}
+	else if (compressionType == COMPRESSION_PG_LZ)
+	{
+		uint32 compressedDataSize = VARSIZE(buffer->data) - CSTORE_COMPRESS_HDRSZ;
+		uint32 decompressedDataSize = CSTORE_COMPRESS_RAWSIZE(buffer->data);
+		char *decompressedData = NULL;
+
+		if (compressedDataSize + CSTORE_COMPRESS_HDRSZ != buffer->len)
+		{
+			ereport(ERROR, (errmsg("cannot decompress the buffer"),
```
onderkalaci: In pg_shard we have a rule that whenever an error message is added, we add a test that hits that line. Is it possible to add a test for this case?

cstore_compression.c
```diff
@@ -0,0 +1,165 @@
...
+
+		int32 decompressedByteCount = 0;
+
+		decompressedByteCount = pglz_decompress(CSTORE_COMPRESS_RAWDATA(buffer->data),
+												compressedDataSize,
+												decompressedData, decompressedDataSize);
+
+		if (decompressedByteCount < 0)
+		{
+			ereport(ERROR, (errmsg("cannot decompress the buffer"),
```
onderkalaci: I know here is trickier to add a tests since it is in `if PG 9.5` block, but if possible could you add a test?

cstore_fdw.c
```diff
@@ -1136,6 +1143,9 @@ CStoreGetForeignPaths(PlannerInfo *root, RelOptInfo *baserel, Oid foreignTableId
 													   startupCost, totalCost,
 													   NIL,  /* no known ordering */
 													   NULL, /* not parameterized */
+#if PG_VERSION_NUM >= 90500
```
onderkalaci: Well, I think it is better to do this in the following pattern:
`If pg 9.5` 
function call with 9.5 API
`else`
function call with 9.4- API
`end if`

An example is on pg_shard as [here](https://github.com/citusdata/pg_shard/pull/158/files#diff-94066e53b61b35da3278923a1fd21124R99). I think that pg_shard's approach is more readable. What do you think?

cstore_fdw.c
```diff
@@ -1177,7 +1194,13 @@ CStoreGetForeignPlan(PlannerInfo *root, RelOptInfo *baserel, Oid foreignTableId,
 	/* create the foreign scan node */
 	foreignScan = make_foreignscan(targetList, scanClauses, baserel->relid,
 								   NIL, /* no expressions to evaluate */
-								   foreignPrivateList);
+								   foreignPrivateList
+#if PG_VERSION_NUM >= 90500
```
onderkalaci: same as above.

cstore_compression.c
mtuncer: We did not follow that approach for cstore_reader.c and cstore_writer.c I don't know the reasoning behind it.

cstore_compression.c
mtuncer: That's more like a 8 byte placeholder. Even structure name is different. I am not sure that requires a copyright message.

cstore_compression.c
mtuncer: only error state is -1 there.  `outputBuffer->data` being null is an error case. I was able to see -1 and > 0  cases so far. Sending an empty buffer (length = 0) also results in -1

cstore_compression.c
mtuncer: that's going to be though here. We need to go corrupt the data file.

cstore_fdw.c
mtuncer: I was divided there. I will update as you suggest.

cstore_fdw.c

cstore_compression.c
onderkalaci: Well, I won't be insisting too much on this. But, I think it'd look better if we have `cstore_compression.h`.

onderkalaci:
It looks good and re-tested after the final commit, it's a :ship: 

(But, I am opening an issue to track moving compression related definitions/declarations into a seperate header with enhancement tag)

PR #79, opened 2016-01-14
release 1.4 changes

mtuncer:


ImreSamu:
Thank you for the release!

Its probably worth mentioning the compatibility with pre-PG96 version in the readme. 
* related to readme : " It doesn't support earlier versions of PostgreSQL versions of PostgreSQL."
* and now available : [PostgreSQL 9.6 Crash test RPM] (http://yum.postgresql.org/repopackages.php#pg96)

PR #87, opened 2016-03-07
Fix copy with transmit option when used in Citus

mtuncer:
cstore_fdw tries to determine if given table is infact a cstore table during copy operation. However,  if given name is not a table at all, cstore fails at this stage and stops copy processing altogether. 

This change  enables cstore_fdw to determine whether a given name is a cstore_fdw table without affecting the rest of copy handling.  If an error is encountered, cstore_fdw stops its own handling only instead of stopping the whole copy operation.

### DIFF
--- a/cstore_fdw.c
+++ b/cstore_fdw.c
@@ -304,7 +304,7 @@ CopyCStoreTableStatement(CopyStmt* copyStatement)
 	if (copyStatement->relation != NULL)
 	{
 		Oid relationId = RangeVarGetRelid(copyStatement->relation,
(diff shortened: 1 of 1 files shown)


sumedhpathak:
Ok, already assigned to @onderkalaci :). @onderkalaci could you review this and get it checked-in tomorrow?  There is some urgency due to a customer request. Thanks!

marcocitus:
Tested with and without Citus. Looks good to me :shipit: 

Is there a way of getting this into PGDG quickly?

PR #89, opened 2016-04-07
Let Citus handle COPY on distributed cstore_fdw tables

marcocitus:
With the new COPY for distributed tables in Citus, we need a way to skip cstore_fdw's COPY implementation for distributed columnar tables. Since citus always needs to come first in shared_preload_libraries, cstore_fdw's ProcessUtility hook always overrides citus' ProcessUtility hook and thus comes first.

After this change, when issuing a COPY command on a columnar table, cstore_fdw first checks whether the Citus extension is loaded, the pg_dist_partition table exists, and  pg_dist_partition contains a record for the cstore_fdw table. If so it lets the COPY fall through to citus.

### DIFF
--- a/cstore_fdw.c
+++ b/cstore_fdw.c
@@ -26,10 +26,12 @@
 #include "access/tuptoaster.h"
 #include "catalog/namespace.h"
 #include "catalog/pg_foreign_table.h"
+#include "catalog/pg_namespace.h"
 #include "commands/copy.h"
 #include "commands/defrem.h"
 #include "commands/event_trigger.h"
 #include "commands/explain.h"
+#include "commands/extension.h"
 #include "commands/vacuum.h"
 #include "foreign/fdwapi.h"
 #include "foreign/foreign.h"
@@ -44,9 +46,11 @@
 #include "parser/parsetree.h"
 #include "tcop/utility.h"
 #include "utils/builtins.h"
+#include "utils/fmgroids.h"
 #include "utils/memutils.h"
 #include "utils/lsyscache.h"
 #include "utils/rel.h"
+#include "utils/tqual.h"
 
 
 /* local functions forward declarations */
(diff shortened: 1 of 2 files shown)


cstore_fdw.c
```diff
@@ -672,6 +681,55 @@ CStoreTable(Oid relationId)
...
+DistributedTable(Oid relationId)
+{
+	bool isDistributedTable = false;
+	Oid partitionOid = InvalidOid;
+	Relation heapRelation = NULL;
+	HeapScanDesc scanDesc = NULL;
+	const int scanKeyCount = 1;
+	ScanKeyData scanKey[scanKeyCount];
+	HeapTuple heapTuple = NULL;
+
```
mtuncer: extra empty line

cstore_fdw.c
```diff
@@ -672,6 +681,55 @@ CStoreTable(Oid relationId)
...
 
 /*
+ * DistributedTable checks if the given relationId is the OID of a distributed table,
+ * which may also be a cstore_fdw table, but in that case COPY should be handled by
+ * Citus.
+ */
+static bool
+DistributedTable(Oid relationId)
+{
+	bool isDistributedTable = false;
```
mtuncer: maybe use distributedTable instead of isDistributedTable ? we do not use -is- prefix in similar cases

mtuncer:
@marcocitus  is there a way to add a test for this change ? Perhaps a fake citus extension and fake pg_dist_partition table.

marcocitus:
@mtuncer Best I could come up with is this:
https://gist.github.com/marcocitus/27b5819895ab6c9c23738f24ae718574

It works both with and without citus in shared_preload_libraries, but it could start crashing on the former if we change Citus in certain ways.

What do you think?

cstore_fdw.c
marcocitus: We usually have an empty line when transitioning from declarations to code (e.g. functions below)

PR #96, opened 2016-05-11
Add new release info

mtuncer:


jasonmp85:
Is there any real reason to bump the version number? I understand making a new release (i.e. maybe `1.4.1`), but as far as PostgreSQL is concerned, do we really need to make the user run an `ALTER EXTENSION UPDATE` if there's no change from the SQL side?

Does this break compatibility with older Citus, or older PostgreSQL? If not, I think we want to bump the release version, but not actually the extension version (i.e. release this as `1.4.1`).

mtuncer:
@jasonmp85 There is no change on sql side. This contains just 2 fixes to support Citus copy function.  We have 2 level version numbers in cstore_fdw, major and minor. Are you saying we still keep it as 1.4 and release as 1.4.1 ?

jasonmp85:
@mtuncer Oh, I see now that this project hasn't used patch versions numbers in the public tags.

For other projects (Citus, Citus Enterprise, pg_shard), we've been using three versions numbers: major, minor, and patch. The SQL extension only uses the first two.

Let's say we release Citus 5.1. The `git` tag would be `5.1.0`, but the SQL version is just `5.1`. Since the SQL version is changing, users would need to update.

But let's say we find a crashing bug in `5.1` and want to fix it quickly. We'd release `5.1.1`, which is a way of saying we aren't changing functionality, just a bug. The `5.1.1` would basically just update the shared library; everything user-facing would look the same.

So if we're not adding functionality, we shouldn't bump the minor or major version, IMO. This feels like a patch fix, _especially_ if it doesn't break compatibility compared with previous releases.

It'll also be easier going forward if we unify our versioning practices. We'll discuss that at the team meeting Monday.

jasonmp85:
I also notice `cstore_fdw` actually uses the major/minor version in serialization code to check for compatibility. We wouldn't need to add patch versions into that check because patch versions should _always_ be compatible with previous releases. Major and minor are sufficient as far as code-compatibility are concerned.

mtuncer:
I am cancelling this PR and opening a new one for 1.4.1

PR #114, opened 2016-07-28
Add alter column type verifications to prevent incorrect changes

### LINKED ISSUE #71: Altering cstore table column data type might fail
Create cstore table, load a row to it:

    create foreign table foo (a int) server cstore_server;
    insert into foo select 1;
    select a from foo;

We get

     a 
    ---
     1
    (1 row)

Now change the column type, this time we are lucky:

    
    alter table foo alter column a type date;
    select a from foo;

And we get

         a      
    ------------
     2000-01-02
    (1 row)

Change the type again, this time we are not lucky:

    alter table foo alter column a type varchar(2);
    select a from foo;

And we get this

    ERROR:  could not open relation with OID 0

And one time, I even got

    ERROR:  invalid memory alloc request size 18446744072304459780

I guess it would be fine to prohibit altering column type to an incompatible one at a first step.

Thanks!

mtuncer:
We relied on PostgreSQL to to type checking and did not verify anything at extension level.
This has caused some usability issues when column type is altered to incompatible type.
We now perform compatibility check for alter table statement before letting standard utility to handle the command.

Fixes #71

### DIFF
--- a/cstore_fdw.c
+++ b/cstore_fdw.c
@@ -44,6 +44,8 @@
 #include "optimizer/var.h"
 #include "parser/parser.h"
 #include "parser/parsetree.h"
+#include "parser/parse_coerce.h"
+#include "parser/parse_type.h"
 #include "tcop/utility.h"
 #include "utils/builtins.h"
 #include "utils/fmgroids.h"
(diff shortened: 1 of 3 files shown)


byucesoy:
This looks good :ship:

PR #117, opened 2016-08-10
Ignore distributed tables in truncate handler

### LINKED ISSUE #116: Truncate handler should ignore distributed tables
cstore_fdw removes columnar tables from truncate command when handling in utility hook. This prevents Citus from truncating distributed cstore_fdw tables.

cstore_fdw should also check if a table is distributed, and ignore that table when handling truncate command.

mtuncer:
Fixes #116


byucesoy:
Should we add test to this? If so please add tests and :ship: or you can :ship: directly if you think we don't need tests.

mtuncer:
unfortunately not at this time. We can mimic cstore's existence by adding dummy pg_dist_partition table. But I think it is a bit overkill.

PR #119, opened 2016-08-31
Remove indirectly dropped cstore table data

### LINKED ISSUE #92: DROP SCHEMA CASCADE doesn't delete cstore files
```
CREATE SCHEMA foo;

CREATE FOREIGN TABLE foo.bar (
  i INT
)
SERVER cstore_server;
```

This creates a `.cstore` and `.cstore.footer` file in the postgres directory, and `DROP FOREIGN TABLE foo.bar;` successfully deletes them again, as discussed in #3.

However, `DROP SCHEMA foo CASCADE;` does not delete these files, and it would be nice if it would. 

Our use case is data warehousing: To ensure an always-consistent set of data, the front-end runs queries against tables in schema `foo`, while ETL scripts progressively build up tables in schema `foo_next`. Once the ETL finishes and data is found to be consistent, we do this:

```
DO $$
BEGIN
  DROP SCHEMA foo CASCADE;
  ALTER SCHEMA foo_next RENAME TO foo;
END
$$;
```

(transactional DDL in Postgres is a super-amazing feature)

### LINKED ISSUE #108: DROP TABLE...CASCADE does not clean up cstore files when run against a parent table
I have a database configuration that includes cstore tables which inherit from a parent table.  When running DROP TABLE...CASCADE against the parent table, the underlying cstore files are not removed from disk.  

 #92 #3

mtuncer:
We were only checking drop foreign table statements when cleaning up cstore data file contents. This approach did not work when a cstore table is dropped as a side effect i.e. when drop schema ... cascade.

This change makes cstore to listen to sql drop trigger and remove dropped cstore files from list acquired from pg_event_trigger_dropped_objects() function.

There are 2 downside to this approach. 
1 - when files are moved they are gone, do chance for rolling back, same as the current behavior.
2 - can not remove data file if filename option is specified while creating the cstore table.  File contents can still be removed by explicitly dropping the table.

fixes #92 
fixes #108


byucesoy:
Looks good :ship:

PR #120, opened 2016-09-05
Bump cstore version to 1.5

mtuncer:


mtuncer:
Also tested with pg96rc1 on ubuntu 14.04

self ⛵

PR #137, opened 2017-02-28
Check whether we're using cstore_fdw server on CREATE FOREIGN TABLE

marcocitus:
cstore_fdw initialises its data files on CREATE FOREIGN TABLE, but it currently does not check whether the table is actually using a cstore_fdw server, which might cause it to overwrite a file when using file_fdw. This change makes sure cstore_fdw only initialise files for cstore_fdw tables.


samay-sharma:
@marcocitus , @mtuncer : Might be worth doing a quick check to confirm that we check this properly for other utility commands in the hooks / event triggers. I recall running into something similar and fixing it while developing a feature. (this was around 3 years back, though :)).

mtuncer:
There was a code duplication in ```CStoreTable``` function. Started to use this function there too.

PR #143, opened 2017-06-01
Use postgres internal storage and enable WAL

mtuncer:
We are going to use postgres storage instead of
regular files. This would be able to create
write logs which will enable replicating
cstore table contents via stream replication.

Filename option that enabled using arbitrary file
path is no longer valid. A new option 'logging' is
introduced to control loggin behavior. Its default
value is set to true meaning logging is enabled.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,8 +17,8 @@ DATA = cstore_fdw--1.5.sql cstore_fdw--1.4--1.5.sql cstore_fdw--1.3--1.4.sql \
 REGRESS = create load query analyze data_types functions block_filtering drop \
 		  insert copyto alter truncate
 EXTRA_CLEAN = cstore.pb-c.h cstore.pb-c.c data/*.cstore data/*.cstore.footer \
-              sql/block_filtering.sql sql/create.sql sql/data_types.sql sql/load.sql \
-              sql/copyto.sql expected/block_filtering.out expected/create.out \
+              sql/block_filtering.sql sql/data_types.sql sql/load.sql \
+			  sql/copyto.sql expected/block_filtering.out \
               expected/data_types.out expected/load.out expected/copyto.out
 
 ifeq ($(enable_coverage),yes)
@@ -32,7 +32,9 @@ endif
 # example: /usr/local/pgsql/bin/pg_config or /usr/lib/postgresql/9.3/bin/pg_config
 #
 
-PG_CONFIG = pg_config
+# PG_CONFIG = pg_config
+PG_CONFIG = /usr/local/pgsql/bin/pg_config
+
 PGXS := $(shell $(PG_CONFIG) --pgxs)
 include $(PGXS)
 
--- a/cstore_fdw--1.5.sql
+++ b/cstore_fdw--1.5.sql
@@ -40,15 +40,15 @@ CREATE OR REPLACE FUNCTION cstore_drop_trigger()
 	RETURNS event_trigger
 	LANGUAGE plpgsql
 	AS $csdt$
-DECLARE v_obj record;
+DECLARE dropped_object record;
 BEGIN
-	FOR v_obj IN SELECT * FROM pg_event_trigger_dropped_objects() LOOP
+	FOR dropped_object IN SELECT * FROM pg_event_trigger_dropped_objects() LOOP
 
-		IF v_obj.object_type NOT IN ('table', 'foreign table') THEN
+		IF dropped_object.object_type <> 'foreign table' THEN
 			CONTINUE;
 		END IF;
 
-		PERFORM cstore_clean_table_resources(v_obj.objid);
+		PERFORM public.cstore_clean_table_resources(dropped_object.objid);
 
 	END LOOP;
 END;
--- a/cstore_fdw.c
+++ b/cstore_fdw.c
@@ -46,6 +46,8 @@
 #include "parser/parsetree.h"
 #include "parser/parse_coerce.h"
 #include "parser/parse_type.h"
+#include "storage/bufmgr.h"
+#include "storage/smgr.h"
 #include "tcop/utility.h"
 #include "utils/builtins.h"
 #include "utils/fmgroids.h"
@@ -54,6 +56,7 @@
 #include "utils/rel.h"
 #include "utils/tqual.h"
 
+#include "catalog/storage.h"
 
 /* local functions forward declarations */
 static void CStoreProcessUtility(Node *parseTree, const char *queryString,
@@ -72,10 +75,8 @@ static uint64 CopyIntoCStoreTable(const CopyStmt *copyStatement,
 								  const char *queryString);
 static uint64 CopyOutCStoreTable(CopyStmt* copyStatement, const char* queryString);
 static void CStoreProcessAlterTableCommand(AlterTableStmt *alterStatement);
-static List * DroppedCStoreFilenameList(DropStmt *dropStatement);
 static List * FindCStoreTables(List *tableList);
 static void TruncateCStoreTables(List *cstoreTableList);
-static void DeleteCStoreTableFiles(char *filename);
 static void InitializeCStoreTableFile(Oid relationId, Relation relation);
 static bool CStoreTable(Oid relationId);
 static bool CStoreServer(ForeignServer *server);
@@ -87,10 +88,10 @@ static void CreateDirectory(StringInfo directoryName);
 static StringInfo OptionNamesString(Oid currentContextId);
 static CStoreFdwOptions * CStoreGetOptions(Oid foreignTableId);
 static char * CStoreGetOptionValue(Oid foreignTableId, const char *optionName);
-static void ValidateForeignTableOptions(char *filename, char *compressionTypeString,
+static void ValidateForeignTableOptions(char *compressionTypeString,
 										char *stripeRowCountString,
-										char *blockRowCountString);
-static char * CStoreDefaultFilePath(Oid foreignTableId);
+										char *blockRowCountString,
+										char *logging);
 static CompressionType ParseCompressionType(const char *compressionTypeString);
 static void CStoreGetForeignRelSize(PlannerInfo *root, RelOptInfo *baserel,
 									Oid foreignTableId);
@@ -106,8 +107,8 @@ static ForeignScan * CStoreGetForeignPlan(PlannerInfo *root, RelOptInfo *baserel
 										  Oid foreignTableId, ForeignPath *bestPath,
 										  List *targetList, List *scanClauses);
 #endif
-static double TupleCountEstimate(RelOptInfo *baserel, const char *filename);
-static BlockNumber PageCount(const char *filename);
+static double TupleCountEstimate(RelOptInfo *baserel, Relation relation);
+static BlockNumber PageCount(Relation relation);
 static List * ColumnList(RelOptInfo *baserel, Oid foreignTableId);
 static void CStoreExplainForeignScan(ForeignScanState *scanState,
 									 ExplainState *explainState);
@@ -212,6 +213,8 @@ cstore_ddl_event_end_trigger(PG_FUNCTION_ARGS)
 			Oid relationId = RangeVarGetRelid(createStatement->base.relation,
 											  AccessShareLock, false);
 			Relation relation = heap_open(relationId, AccessExclusiveLock);
+			RelationCreateStorage(relation->rd_node, RELPERSISTENCE_PERMANENT);
+
 			InitializeCStoreTableFile(relationId, relation);
 			heap_close(relation, AccessExclusiveLock);
 		}
@@ -245,22 +248,6 @@ CStoreProcessUtility(Node *parseTree, const char *queryString,
 			CallPreviousProcessUtility(parseTree, queryString, context,
 									   paramListInfo, destReceiver, completionTag);
 		}
-	}
-	else if (nodeTag(parseTree) == T_DropStmt)
-	{
-		ListCell *fileListCell = NULL;
-		List *droppedTables = DroppedCStoreFilenameList((DropStmt*) parseTree);
-
-		CallPreviousProcessUtility(parseTree, queryString, context,
-								   paramListInfo, destReceiver, completionTag);
-
-		foreach(fileListCell, droppedTables)
-		{
-			char *fileName = lfirst(fileListCell);
-
-			DeleteCStoreTableFiles(fileName);
-		}
-
 	}
 	else if (nodeTag(parseTree) == T_TruncateStmt)
 	{
@@ -472,10 +459,10 @@ CopyIntoCStoreTable(const CopyStmt *copyStatement, const char *queryString)
 							  copyStatement->options);
 
 	/* init state to write to the cstore file */
-	writeState = CStoreBeginWrite(cstoreFdwOptions->filename,
-								  cstoreFdwOptions->compressionType,
+	writeState = CStoreBeginWrite(relation, cstoreFdwOptions->compressionType,
 								  cstoreFdwOptions->stripeRowCount,
 								  cstoreFdwOptions->blockRowCount,
+								  cstoreFdwOptions->logging,
 								  tupleDescriptor);
 
 	while (nextRowFound)
@@ -610,49 +597,6 @@ CStoreProcessAlterTableCommand(AlterTableStmt *alterStatement)
 }
 
 
-/*
- * DropppedCStoreFilenameList extracts and returns the list of cstore file names
- * from DROP table statement
- */
-static List *
-DroppedCStoreFilenameList(DropStmt *dropStatement)
-{
-	List *droppedCStoreFileList = NIL;
-
-	if (dropStatement->removeType == OBJECT_FOREIGN_TABLE)
-	{
-		ListCell *dropObjectCell = NULL;
-		foreach(dropObjectCell, dropStatement->objects)
-		{
-			List *tableNameList = (List *) lfirst(dropObjectCell);
-			RangeVar *rangeVar = makeRangeVarFromNameList(tableNameList);
-
-			Oid relationId = RangeVarGetRelid(rangeVar, AccessShareLock, true);
-			if (CStoreTable(relationId))
-			{
-				CStoreFdwOptions *cstoreFdwOptions = CStoreGetOptions(relationId);
-				char *defaultfilename = CStoreDefaultFilePath(relationId);
-
-				/*
-				 * Skip files that are placed in default location, they are handled
-				 * by sql drop trigger. Both paths are generated by code, use
-				 * of strcmp is safe here.
-				 */
-				if (strcmp(defaultfilename, cstoreFdwOptions->filename) == 0)
-				{
-					continue;
-				}
-
-				droppedCStoreFileList = lappend(droppedCStoreFileList,
-												cstoreFdwOptions->filename);
-			}
-		}
-	}
-
-	return droppedCStoreFileList;
-}
-
-
 /* FindCStoreTables returns list of CStore tables from given table list */
 static List *
 FindCStoreTables(List *tableList)
@@ -683,12 +627,14 @@ TruncateCStoreTables(List *cstoreTableList)
 		RangeVar *rangeVar = (RangeVar *) lfirst(relationCell);
 		Oid relationId = RangeVarGetRelid(rangeVar, AccessShareLock, true);
 		Relation relation = NULL;
-		CStoreFdwOptions *cstoreFdwOptions = NULL;
-
+		SMgrRelation smgrRelation = NULL;
 		Assert(CStoreTable(relationId));
 		relation = heap_open(relationId, AccessExclusiveLock);
-		cstoreFdwOptions = CStoreGetOptions(relationId);
-		DeleteCStoreTableFiles(cstoreFdwOptions->filename);
+
+		smgrRelation = smgropen(relation->rd_node, InvalidBackendId);
+		smgrtruncate(smgrRelation, FOOTER_FORKNUM, 0);
+		smgrtruncate(smgrRelation, DATA_FORKNUM, 0);
+
 		InitializeCStoreTableFile(relationId, relation);
 		heap_close(relation, AccessExclusiveLock);
 	}
@@ -696,43 +642,8 @@ TruncateCStoreTables(List *cstoreTableList)
 
 
 /*
- * DeleteCStoreTableFiles deletes the data and footer files for a cstore table
- * whose data filename is given.
- */
-static void
-DeleteCStoreTableFiles(char *filename)
-{
-	int dataFileRemoved = 0;
-	int footerFileRemoved = 0;
-
-	StringInfo tableFooterFilename = makeStringInfo();
-	appendStringInfo(tableFooterFilename, "%s%s", filename, CSTORE_FOOTER_FILE_SUFFIX);
-
-	/* delete the footer file */
-	footerFileRemoved = unlink(tableFooterFilename->data);
-	if (footerFileRemoved != 0)
-	{
-		ereport(WARNING, (errcode_for_file_access(),
-						  errmsg("could not delete file \"%s\": %m",
-								 tableFooterFilename->data)));
-	}
-
-	/* delete the data file */
-	dataFileRemoved = unlink(filename);
-	if (dataFileRemoved != 0)
-	{
-		ereport(WARNING, (errcode_for_file_access(),
-						  errmsg("could not delete file \"%s\": %m",
-								 filename)));
-	}
-}
-
-
-/*
- * InitializeCStoreTableFile creates data and footer file for a cstore table.
- * The function assumes data and footer files do not exist, therefore
- * it should be called on empty or non-existing table. Notice that the caller
- * is expected to acquire AccessExclusiveLock on the relation.
+ * InitializeCStoreTableFile creates an initializes data and footer files
+ * for a cstore table. Files are allocated in postgres internal storage.
  */
 static void InitializeCStoreTableFile(Oid relationId, Relation relation)
 {
@@ -744,9 +655,13 @@ static void InitializeCStoreTableFile(Oid relationId, Relation relation)
 	 * Initialize state to write to the cstore file. This creates an
 	 * empty data file and a valid footer file for the table.
 	 */
-	writeState = CStoreBeginWrite(cstoreFdwOptions->filename,
-			cstoreFdwOptions->compressionType, cstoreFdwOptions->stripeRowCount,
-			cstoreFdwOptions->blockRowCount, tupleDescriptor);
+	writeState = CStoreBeginWrite(relation,
+								  cstoreFdwOptions->compressionType,
+								  cstoreFdwOptions->stripeRowCount,
+								  cstoreFdwOptions->blockRowCount,
+								  cstoreFdwOptions->logging,
+								  tupleDescriptor);
+
 	CStoreEndWrite(writeState);
 }
 
@@ -956,7 +871,7 @@ CreateDirectory(StringInfo directoryName)
 
 /*
  * cstore_table_size returns the total on-disk size of a cstore table in bytes.
- * The result includes the sizes of data file and footer file.
+ * The result includes the sizes of data and footer.
  */
 Datum
 cstore_table_size(PG_FUNCTION_ARGS)
@@ -964,13 +879,9 @@ cstore_table_size(PG_FUNCTION_ARGS)
 	Oid relationId = PG_GETARG_OID(0);
 
 	int64 tableSize = 0;
-	CStoreFdwOptions *cstoreFdwOptions = NULL;
-	char *dataFilename = NULL;
-	StringInfo footerFilename = NULL;
-	int dataFileStatResult = 0;
-	int footerFileStatResult = 0;
-	struct stat dataFileStatBuffer;
-	struct stat footerFileStatBuffer;
+	BlockNumber pageCount = 0;
+
+	Relation relation = NULL;
 
 	bool cstoreTable = CStoreTable(relationId);
 	if (!cstoreTable)
@@ -978,30 +889,11 @@ cstore_table_size(PG_FUNCTION_ARGS)
 		ereport(ERROR, (errmsg("relation is not a cstore table")));
 	}
 
-	cstoreFdwOptions = CStoreGetOptions(relationId);
-	dataFilename = cstoreFdwOptions->filename;
+	relation = heap_open(relationId, AccessExclusiveLock);
+	pageCount = PageCount(relation);
+	heap_close(relation, AccessExclusiveLock);
 
-	dataFileStatResult = stat(dataFilename, &dataFileStatBuffer);
-	if (dataFileStatResult != 0)
-	{
-		ereport(ERROR, (errcode_for_file_access(),
-						errmsg("could not stat file \"%s\": %m", dataFilename)));
-	}
-
-	footerFilename = makeStringInfo();
-	appendStringInfo(footerFilename, "%s%s", dataFilename,
-					 CSTORE_FOOTER_FILE_SUFFIX);
-
-	footerFileStatResult = stat(footerFilename->data, &footerFileStatBuffer);
-	if (footerFileStatResult != 0)
-	{
-		ereport(ERROR, (errcode_for_file_access(),
-						errmsg("could not stat file \"%s\": %m",
-								footerFilename->data)));
-	}
-
-	tableSize += dataFileStatBuffer.st_size;
-	tableSize += footerFileStatBuffer.st_size;
+	tableSize = pageCount * BLCKSZ;
 
 	PG_RETURN_INT64(tableSize);
 }
@@ -1046,10 +938,10 @@ cstore_fdw_validator(PG_FUNCTION_ARGS)
 	Oid optionContextId = PG_GETARG_OID(1);
 	List *optionList = untransformRelOptions(optionArray);
 	ListCell *optionCell = NULL;
-	char *filename = NULL;
 	char *compressionTypeString = NULL;
 	char *stripeRowCountString = NULL;
 	char *blockRowCountString = NULL;
+	char *loggingString = NULL;
 
 	foreach(optionCell, optionList)
 	{
@@ -1081,11 +973,7 @@ cstore_fdw_validator(PG_FUNCTION_ARGS)
 									optionNamesString->data)));
 		}
 
-		if (strncmp(optionName, OPTION_NAME_FILENAME, NAMEDATALEN) == 0)
-		{
-			filename = defGetString(optionDef);
-		}
-		else if (strncmp(optionName, OPTION_NAME_COMPRESSION_TYPE, NAMEDATALEN) == 0)
+		if (strncmp(optionName, OPTION_NAME_COMPRESSION_TYPE, NAMEDATALEN) == 0)
 		{
 			compressionTypeString = defGetString(optionDef);
 		}
@@ -1097,12 +985,17 @@ cstore_fdw_validator(PG_FUNCTION_ARGS)
 		{
 			blockRowCountString = defGetString(optionDef);
 		}
+		else if (strncmp(optionName, OPTION_NAME_LOGGING, NAMEDATALEN) == 0)
+		{
+			loggingString = defGetString(optionDef);
+		}
 	}
 
 	if (optionContextId == ForeignTableRelationId)
 	{
-		ValidateForeignTableOptions(filename, compressionTypeString,
-									stripeRowCountString, blockRowCountString);
+		ValidateForeignTableOptions(compressionTypeString,
+									stripeRowCountString, blockRowCountString,
+									loggingString);
 	}
 
 	PG_RETURN_VOID();
@@ -1121,22 +1014,12 @@ Datum
 cstore_clean_table_resources(PG_FUNCTION_ARGS)
 {
 	Oid relationId = PG_GETARG_OID(0);
-	StringInfo filePath = makeStringInfo();
-	struct stat fileStat;
-	int statResult = -1;
+	RelFileNode relFileNode = {MyDatabaseTableSpace, MyDatabaseId, relationId};
+	Relation relation = palloc0(sizeof(RelationData));
+	relation->rd_node = relFileNode;
+	relation->rd_backend = InvalidBackendId;
 
-	appendStringInfo(filePath, "%s/%s/%d/%d", DataDir, CSTORE_FDW_NAME,
-					 (int) MyDatabaseId, (int) relationId);
-
-	/*
-	 * Check to see if the file exist first. This is the only way to
-	 * find out if the table being dropped is a cstore table.
-	 */
-	statResult = stat(filePath->data, &fileStat);
-	if (statResult == 0)
-	{
-		DeleteCStoreTableFiles(filePath->data);
-	}
+	RelationDropStorage(relation);
 
 	PG_RETURN_VOID();
 }
@@ -1185,15 +1068,16 @@ static CStoreFdwOptions *
 CStoreGetOptions(Oid foreignTableId)
 {
 	CStoreFdwOptions *cstoreFdwOptions = NULL;
-	char *filename = NULL;
+
 	CompressionType compressionType = DEFAULT_COMPRESSION_TYPE;
 	int32 stripeRowCount = DEFAULT_STRIPE_ROW_COUNT;
 	int32 blockRowCount = DEFAULT_BLOCK_ROW_COUNT;
+	bool logging = true;
 	char *compressionTypeString = NULL;
 	char *stripeRowCountString = NULL;
 	char *blockRowCountString = NULL;
+	char *loggingString = NULL;
 
-	filename = CStoreGetOptionValue(foreignTableId, OPTION_NAME_FILENAME);
 	compressionTypeString = CStoreGetOptionValue(foreignTableId,
 												 OPTION_NAME_COMPRESSION_TYPE);
 	stripeRowCountString = CStoreGetOptionValue(foreignTableId,
@@ -1201,8 +1085,11 @@ CStoreGetOptions(Oid foreignTableId)
 	blockRowCountString = CStoreGetOptionValue(foreignTableId,
 											   OPTION_NAME_BLOCK_ROW_COUNT);
 
-	ValidateForeignTableOptions(filename, compressionTypeString,
-								stripeRowCountString, blockRowCountString);
+	loggingString = CStoreGetOptionValue(foreignTableId,
+										 OPTION_NAME_LOGGING);
+
+	ValidateForeignTableOptions(compressionTypeString,
+								stripeRowCountString, blockRowCountString, loggingString);
 
 	/* parse provided options */
 	if (compressionTypeString != NULL)
@@ -1217,18 +1104,16 @@ CStoreGetOptions(Oid foreignTableId)
 	{
 		blockRowCount = pg_atoi(blockRowCountString, sizeof(int32), 0);
 	}
-
-	/* set default filename if it is not provided */
-	if (filename == NULL)
+	if (loggingString != NULL)
 	{
-		filename = CStoreDefaultFilePath(foreignTableId);
+		logging = strncmp(loggingString, LOGGING_STRING_TRUE, NAMEDATALEN) == 0;
 	}
 
 	cstoreFdwOptions = palloc0(sizeof(CStoreFdwOptions));
-	cstoreFdwOptions->filename = filename;
 	cstoreFdwOptions->compressionType = compressionType;
 	cstoreFdwOptions->stripeRowCount = stripeRowCount;
 	cstoreFdwOptions->blockRowCount = blockRowCount;
+	cstoreFdwOptions->logging = logging;
 
 	return cstoreFdwOptions;
 }
@@ -1276,12 +1161,10 @@ CStoreGetOptionValue(Oid foreignTableId, const char *optionName)
  * considered invalid.
  */
 static void
-ValidateForeignTableOptions(char *filename, char *compressionTypeString,
-							char *stripeRowCountString, char *blockRowCountString)
+ValidateForeignTableOptions(char *compressionTypeString,
+							char *stripeRowCountString, char *blockRowCountString,
+							char *loggingString)
 {
-	/* we currently do not have any checks for filename */
-	(void) filename;
-
 	/* check if the provided compression type is valid */
 	if (compressionTypeString != NULL)
 	{
@@ -1323,29 +1206,18 @@ ValidateForeignTableOptions(char *filename, char *compressionTypeString,
 									BLOCK_ROW_COUNT_MAXIMUM)));
 		}
 	}
-}
 
-
-/*
- * CStoreDefaultFilePath constructs the default file path to use for a cstore_fdw
- * table. The path is of the form $PGDATA/cstore_fdw/{databaseOid}/{relfilenode}.
- */
-static char *
-CStoreDefaultFilePath(Oid foreignTableId)
-{
-	Relation relation = relation_open(foreignTableId, AccessShareLock);
-	RelFileNode relationFileNode = relation->rd_node; 
-
-	Oid databaseOid = relationFileNode.dbNode;
-	Oid relationFileOid = relationFileNode.relNode;
-
-	StringInfo cstoreFilePath = makeStringInfo();
-	appendStringInfo(cstoreFilePath, "%s/%s/%u/%u", DataDir, CSTORE_FDW_NAME,
-					 databaseOid, relationFileOid);
-
-	relation_close(relation, AccessShareLock);
-
-	return cstoreFilePath->data;
+	/* check if the provided logging flag is valid */
+	if (loggingString != NULL)
+	{
+		if (strncmp(LOGGING_STRING_TRUE, loggingString, NAMEDATALEN) != 0 &&
+			strncmp(LOGGING_STRING_FALSE, loggingString, NAMEDATALEN) != 0)
+		{
+			ereport(ERROR, (errmsg("invalid logging flag"),
+							errhint("Valid options are: %s",
+									LOGGING_STRING_DELIMITED_LIST)));
+		}
+	}
 }
 
 
@@ -1376,13 +1248,14 @@ ParseCompressionType(const char *compressionTypeString)
 static void
 CStoreGetForeignRelSize(PlannerInfo *root, RelOptInfo *baserel, Oid foreignTableId)
 {
-	CStoreFdwOptions *cstoreFdwOptions = CStoreGetOptions(foreignTableId);
-	double tupleCountEstimate = TupleCountEstimate(baserel, cstoreFdwOptions->filename);
+	Relation relation = heap_open(foreignTableId, AccessShareLock);
+	double tupleCountEstimate = TupleCountEstimate(baserel, relation);
 	double rowSelectivity = clauselist_selectivity(root, baserel->baserestrictinfo,
 												   0, JOIN_INNER, NULL);
 
 	double outputRowCount = clamp_row_est(tupleCountEstimate * rowSelectivity);
 	baserel->rows = outputRowCount;
+	heap_close(relation, AccessShareLock);
 }
 
 
@@ -1396,7 +1269,6 @@ static void
 CStoreGetForeignPaths(PlannerInfo *root, RelOptInfo *baserel, Oid foreignTableId)
 {
 	Path *foreignScanPath = NULL;
-	CStoreFdwOptions *cstoreFdwOptions = CStoreGetOptions(foreignTableId);
 	Relation relation = heap_open(foreignTableId, AccessShareLock);
 
 	/*
@@ -1417,14 +1289,14 @@ CStoreGetForeignPaths(PlannerInfo *root, RelOptInfo *baserel, Oid foreignTableId
 	 */
 	List *queryColumnList = ColumnList(baserel, foreignTableId);
 	uint32 queryColumnCount = list_length(queryColumnList);
-	BlockNumber relationPageCount = PageCount(cstoreFdwOptions->filename);
+	BlockNumber relationPageCount = PageCount(relation);
 	uint32 relationColumnCount = RelationGetNumberOfAttributes(relation);
 
 	double queryColumnRatio = (double) queryColumnCount / relationColumnCount;
 	double queryPageCount = relationPageCount * queryColumnRatio;
 	double totalDiskAccessCost = seq_page_cost * queryPageCount;
 
-	double tupleCountEstimate = TupleCountEstimate(baserel, cstoreFdwOptions->filename);
+	double tupleCountEstimate = TupleCountEstimate(baserel, relation);
 
 	/*
 	 * We estimate costs almost the same way as cost_seqscan(), thus assuming
@@ -1529,7 +1401,7 @@ CStoreGetForeignPlan(PlannerInfo *root, RelOptInfo *baserel, Oid foreignTableId,
  * file.
  */
 static double
-TupleCountEstimate(RelOptInfo *baserel, const char *filename)
+TupleCountEstimate(RelOptInfo *baserel, Relation relation)
 {
 	double tupleCountEstimate = 0.0;
 
@@ -1542,13 +1414,13 @@ TupleCountEstimate(RelOptInfo *baserel, const char *filename)
 		 * that by the current file size.
 		 */
 		double tupleDensity = baserel->tuples / (double) baserel->pages;
-		BlockNumber pageCount = PageCount(filename);
+		BlockNumber pageCount = PageCount(relation);
 
 		tupleCountEstimate = clamp_row_est(tupleDensity * (double) pageCount);
 	}
 	else
 	{
-		tupleCountEstimate = (double) CStoreTableRowCount(filename);
+		tupleCountEstimate = (double) CStoreTableRowCount(relation);
 	}
 
 	return tupleCountEstimate;
@@ -1557,22 +1429,21 @@ TupleCountEstimate(RelOptInfo *baserel, const char *filename)
 
 /* PageCount calculates and returns the number of pages in a file. */
 static BlockNumber
-PageCount(const char *filename)
+PageCount(Relation relation)
 {
 	BlockNumber pageCount = 0;
-	struct stat statBuffer;
+	BlockNumber footerBlockCount = 0;
+	BlockNumber dataBlockCount = 0;
 
-	/* if file doesn't exist at plan time, use default estimate for its size */
-	int statResult = stat(filename, &statBuffer);
-	if (statResult < 0)
-	{
-		statBuffer.st_size = 10 * BLCKSZ;
-	}
+	footerBlockCount = RelationGetNumberOfBlocksInFork(relation, FOOTER_FORKNUM);
+	dataBlockCount = RelationGetNumberOfBlocksInFork(relation, DATA_FORKNUM);
 
-	pageCount = (statBuffer.st_size + (BLCKSZ - 1)) / BLCKSZ;
-	if (pageCount < 1)
+	pageCount = footerBlockCount + dataBlockCount;
+
+	/* if the relation is empty, use default estimate for its size */
+	if (pageCount == 0)
 	{
-		pageCount = 1;
+		pageCount = 10;
 	}
 
 	return pageCount;
@@ -1689,22 +1560,17 @@ ColumnList(RelOptInfo *baserel, Oid foreignTableId)
 static void
 CStoreExplainForeignScan(ForeignScanState *scanState, ExplainState *explainState)
 {
-	Oid foreignTableId = RelationGetRelid(scanState->ss.ss_currentRelation);
-	CStoreFdwOptions *cstoreFdwOptions = CStoreGetOptions(foreignTableId);
-
-	ExplainPropertyText("CStore File", cstoreFdwOptions->filename, explainState);
-
 	/* supress file size if we're not showing cost details */
 	if (explainState->costs)
 	{
-		struct stat statBuffer;
+		Oid foreignTableId = RelationGetRelid(scanState->ss.ss_currentRelation);
+		Relation relation = heap_open(foreignTableId, AccessExclusiveLock);
+		BlockNumber pageCount = PageCount(relation);
+		long fileSize = pageCount * BLCKSZ;
 
-		int statResult = stat(cstoreFdwOptions->filename, &statBuffer);
-		if (statResult == 0)
-		{
-			ExplainPropertyLong("CStore File Size", (long) statBuffer.st_size,
-								explainState);
-		}
+		heap_close(relation, AccessExclusiveLock);
+
+		ExplainPropertyLong("CStore File Size", fileSize, explainState);
 	}
 }
 
@@ -1722,6 +1588,7 @@ CStoreBeginForeignScan(ForeignScanState *scanState, int executorFlags)
 	ForeignScan *foreignScan = NULL;
 	List *foreignPrivateList = NIL;
 	List *whereClauseList = NIL;
+	Relation currentRelation = scanState->ss.ss_currentRelation;
 
 	/* if Explain with no Analyze, do nothing */
 	if (executorFlags & EXEC_FLAG_EXPLAIN_ONLY)
@@ -1729,7 +1596,7 @@ CStoreBeginForeignScan(ForeignScanState *scanState, int executorFlags)
 		return;
 	}
 
-	foreignTableId = RelationGetRelid(scanState->ss.ss_currentRelation);
+	foreignTableId = RelationGetRelid(currentRelation);
 	cstoreFdwOptions = CStoreGetOptions(foreignTableId);
 
 	foreignScan = (ForeignScan *) scanState->ss.ps.plan;
@@ -1737,8 +1604,8 @@ CStoreBeginForeignScan(ForeignScanState *scanState, int executorFlags)
 	whereClauseList = foreignScan->scan.plan.qual;
 
 	columnList = (List *) linitial(foreignPrivateList);
-	readState = CStoreBeginRead(cstoreFdwOptions->filename, tupleDescriptor,
-								columnList, whereClauseList);
+	readState = CStoreBeginRead(currentRelation, tupleDescriptor, columnList,
+								whereClauseList);
 
 	scanState->fdw_state = (void *) readState;
 }
@@ -1807,19 +1674,7 @@ CStoreAnalyzeForeignTable(Relation relation,
 						  AcquireSampleRowsFunc *acquireSampleRowsFunc,
 						  BlockNumber *totalPageCount)
 {
-	Oid foreignTableId = RelationGetRelid(relation);
-	CStoreFdwOptions *cstoreFdwOptions = CStoreGetOptions(foreignTableId);
-	struct stat statBuffer;
-
-	int statResult = stat(cstoreFdwOptions->filename, &statBuffer);
-	if (statResult < 0)
-	{
-		ereport(ERROR, (errcode_for_file_access(),
-						errmsg("could not stat file \"%s\": %m",
-							   cstoreFdwOptions->filename)));
-	}
-
-	(*totalPageCount) = PageCount(cstoreFdwOptions->filename);
+	(*totalPageCount) = PageCount(relation);
 	(*acquireSampleRowsFunc) = CStoreAcquireSampleRows;
 
 	return true;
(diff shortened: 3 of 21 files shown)


cstore_fdw.c
```diff
@@ -235,17 +226,12 @@ cstore_ddl_event_end_trigger(PG_FUNCTION_ARGS)
...
 
-			/*
-			 * Make sure database directory exists before creating a table.
-			 * This is necessary when a foreign server is created inside
-			 * a template database and a new database is created out of it.
-			 * We have no chance to hook into server creation to create data
-			 * directory for it during database creation time.
-			 */
-			CreateCStoreDatabaseDirectory(MyDatabaseId);
+			RelationCreateStorage(relation->rd_node, RELPERSISTENCE_PERMANENT);
```
anarazel: Seems better to use `relation->rd_rel->relpersistence`.

cstore_fdw.c
```diff
@@ -254,6 +240,53 @@ cstore_ddl_event_end_trigger(PG_FUNCTION_ARGS)
...
+	}
+	else if (functionList->next != NULL)
+	{
+		ereport(ERROR, (errcode(ERRCODE_AMBIGUOUS_FUNCTION),
+						errmsg("more than one function named \"%s\"", functionName)));
+	}
+
+	functionOid = functionList->oid;
+
+	OidFunctionCall1(functionOid, ObjectIdGetDatum(relationId));
```
anarazel: I'd rather do a `DirectFunctionCall(register_cstore_table, ObjectIdGetDatum(relationId))` - that saves all the lookup rigamarole.

cstore_fdw.c
```diff
@@ -317,33 +351,33 @@ CStoreProcessUtility(Node * parseTree, const char *queryString,
...
-			ListCell *fileListCell = NULL;
-			List *droppedTables = DroppedCStoreFilenameList((DropStmt *) parseTree);
 
-			CALL_PREVIOUS_UTILITY(parseTree, queryString, context, paramListInfo,
-								  destReceiver, completionTag);
+			CALL_PREVIOUS_UTILITY(parseTree, queryString, context,
+					paramListInfo, destReceiver, completionTag);
 
-			foreach(fileListCell, droppedTables)
+			if (cstoreTableList != NIL)
```
anarazel: What's this if for?

cstore_fdw.c
```diff
@@ -733,46 +752,58 @@ CStoreProcessAlterTableCommand(AlterTableStmt *alterStatement)
...
 
 
-/*
- * DropppedCStoreFilenameList extracts and returns the list of cstore file names
- * from DROP table statement
- */
+/* CStoreTableList returns oids of cstore_fdw tables stored in pg_cstore_tables */
 static List *
-DroppedCStoreFilenameList(DropStmt *dropStatement)
+CStoreTableList()
```
anarazel: Missing `void` in params.

cstore_fdw.c
```diff
@@ -733,46 +752,58 @@ CStoreProcessAlterTableCommand(AlterTableStmt *alterStatement)
...
+		Oid relationId = DatumGetObjectId(logicalrelidDatum);
+
+		cstoreRelationIdList = lappend_oid(cstoreRelationIdList, relationId);
+
+		heapTuple = heap_getnext(scanDesc, ForwardScanDirection);
 	}
 
-	return droppedCStoreFileList;
+	heap_endscan(scanDesc);
+	heap_close(heapRelation, AccessShareLock);
```
anarazel: Why are you dropping the lock here?

cstore_fdw.c
```diff
@@ -733,46 +752,58 @@ CStoreProcessAlterTableCommand(AlterTableStmt *alterStatement)
...
+
+	return cstoreRelationIdList;
+}
+
+
+ /* RemoveRelationStorage marks storage nodes for a relation for deletion */
+static void
+RemoveRelationStorage(Oid relationId)
+{
+	RelFileNode relFileNode = {MyDatabaseTableSpace, MyDatabaseId, relationId};
```
anarazel: Are you sure that relationId is always the correct thing here? Even after e.g. a trunction?

cstore_metadata_serialization.c
```diff
@@ -579,3 +580,77 @@ ProtobufBinaryToDatum(ProtobufCBinaryData protobufBinary, bool datumTypeByValue,
...
+ * starts and length of footer data in number of pages.
+ * Expects to see cstore magic number (char[12]), major version (uint32), minor version(uint32)
+ * startingPage (uint32), pageCount (uint32)
+ */
+bool
+DeserializeTableFooterMetadata(const char *data, uint32 length, uint32 *startingPage,
+							   uint32 *pageCount)
+{
+	uint32 offset = 0;
+	char *magicNumber = pstrdup(CSTORE_MAGIC_NUMBER);
```
anarazel: What's this pstrdup achieving?

cstore_metadata_serialization.c
```diff
@@ -579,3 +580,77 @@ ProtobufBinaryToDatum(ProtobufCBinaryData protobufBinary, bool datumTypeByValue,
...
+	memcpy(&majorVersion, data + offset, sizeof(uint32));
+	offset += sizeof(uint32);
+
+	memcpy(&minorVersion, data + offset, sizeof(uint32));
+	offset += sizeof(uint32);
+
+	if (majorVersion != CSTORE_VERSION_MAJOR ||
+		minorVersion != CSTORE_VERSION_MINOR)
+	{
+		ereport(ERROR, (errmsg("Unsupported cstore_fdw version")));
```
anarazel: I'd expand this error to say that the relation has an unexpected cstore version. And also error messages start with a lower case...

cstore_metadata_serialization.c
```diff
@@ -579,3 +580,77 @@ ProtobufBinaryToDatum(ProtobufCBinaryData protobufBinary, bool datumTypeByValue,
...
+{
+	StringInfo resultString = makeStringInfo();
+	uint32 majorVersion = CSTORE_VERSION_MAJOR;
+	uint32 minorVersion = CSTORE_VERSION_MINOR;
+
+	appendBinaryStringInfo(resultString, CSTORE_MAGIC_NUMBER,
+						   strnlen(CSTORE_MAGIC_NUMBER, NAMEDATALEN));
+	appendBinaryStringInfo(resultString, (char *) &majorVersion, sizeof(uint32));
+	appendBinaryStringInfo(resultString, (char *) &minorVersion, sizeof(uint32));
+	appendBinaryStringInfo(resultString, (char *)  &startingPage, sizeof(uint32));
```
anarazel: weird spaces.

cstore_reader.c
```diff
@@ -147,78 +136,166 @@ CStoreBeginRead(const char *filename, TupleDesc tupleDescriptor,
...
+	return tableFooter;
+}
+
+
+/*
+ * ReadFooterData reads and returns footer data. The first page contains the
+ * metadata about footer data. The function validates the metadata, reads
+ * footer data from where it is stored and returns it.
+ */
+static StringInfo ReadFooterData(Relation relation)
```
anarazel: Formatting.

cstore_reader.c
```diff
@@ -147,78 +136,166 @@ CStoreBeginRead(const char *filename, TupleDesc tupleDescriptor,
...
+		pageHeader = (PageHeader) page;
+		pageData = PageGetContents(page);
+		pageDataLength = pageHeader->pd_lower - SizeOfPageHeaderData;
+		appendBinaryStringInfo(footerData, pageData, pageDataLength);
+		LockBuffer(buffer, BUFFER_LOCK_UNLOCK);
+		ReleaseBuffer(buffer);
+	}
+
+	UnlockReleaseBuffer(firstBuffer);
+	UnlockReleaseBuffer(headerBuffer);
```
anarazel: Why are we keeping both of these locked?

cstore_reader.c
```diff
@@ -1287,76 +1354,78 @@ ColumnDefaultValue(TupleConstr *tupleConstraints, Form_pg_attribute attributeFor
...
+	uint32 blockOffset = offset % blockCapacity;
+	uint32 remainingSize = size;
+	uint32 resultOffset = 0;
+	uint32 blockCount = 0;
+	StringInfo resultBuffer = NULL;
 
-	errno = 0;
-	fseekResult = fseeko(file, 0, SEEK_END);
-	if (fseekResult != 0)
+	if (size == 0)
```
anarazel: Gah, this is one of these cases where returning a variable is a lot less clear than just returning `NULL`.

cstore_reader.c
```diff
@@ -1287,76 +1354,78 @@ ColumnDefaultValue(TupleConstr *tupleConstraints, Form_pg_attribute attributeFor
...
+	{
+		Buffer buffer = ReadBufferExtended(relation, DATA_FORKNUM, blockNumber,
+										   RBM_NORMAL, NULL);
+		Page page = NULL;
+		Size pageSize = 0;
+		char *pageContents = NULL;
+		PageHeader pageHeader = NULL;
+
+		if (buffer == InvalidBuffer)
+		{
```
anarazel: I don't see how this is a reachable path? To wit:
```
 *
 * Returns: the buffer number for the buffer containing
 *		the block read.  The returned buffer has been pinned.
 *		Does not return on error --- elog's instead.
 *
```

cstore_reader.c
```diff
@@ -1287,76 +1354,78 @@ ColumnDefaultValue(TupleConstr *tupleConstraints, Form_pg_attribute attributeFor
...
-	int fseekResult = 0;
-	int freadResult = 0;
-	int fileError = 0;
+		LockBuffer(buffer, BUFFER_LOCK_SHARE);
+		page = BufferGetPage(buffer);
+		pageHeader = (PageHeader ) page;
+		pageSize = PageGetPageSize(page);
+		pageContents = PageGetContents(page);
+		uint32 bufferSize = pageHeader->pd_lower - SizeOfPageHeaderData;
+		uint32 copySize = remainingSize;
```
anarazel: Something in the indentation is screwy here.

cstore_writer.c
```diff
@@ -191,12 +174,37 @@ CStoreBeginWrite(const char *filename, CompressionType compressionType,
...
+ */
+static void EnsureDataForkExists(Relation relation, bool logging)
+{
+	SMgrRelation srel = smgropen(relation->rd_node, InvalidBackendId);
+
+	if (!smgrexists(srel, DATA_FORKNUM))
+	{
+		smgrcreate(srel, DATA_FORKNUM, false);
+
+		if (logging)
```
anarazel: Perhaps add a comment that this function requires an exclusive lock?

cstore_writer.c
```diff
@@ -191,12 +174,37 @@ CStoreBeginWrite(const char *filename, CompressionType compressionType,
...
+		{
+			log_smgrcreate(&srel->smgr_rnode.node, DATA_FORKNUM);
```
anarazel: Why are passing through a `logging` option here? Couldn't we just use `RelationNeedsWAL()`?

cstore_writer.c
```diff
@@ -191,12 +174,37 @@ CStoreBeginWrite(const char *filename, CompressionType compressionType,
...
 }
 
 
 /*
+ * EnsureDataForkExists checks if data fork exists for writing data,
+ * and creates if it is not present.
+ */
+static void EnsureDataForkExists(Relation relation, bool logging)
+{
+	SMgrRelation srel = smgropen(relation->rd_node, InvalidBackendId);
```
anarazel: Should use `RelationOpenSmgr()` (everywhere).

cstore_writer.c
```diff
@@ -318,96 +326,221 @@ CStoreWriteRow(TableWriteState *writeState, Datum *columnValues, bool *columnNul
...
 	pfree(tableFooterBuffer);
 	pfree(postscriptBuffer->data);
 	pfree(postscriptBuffer);
+
+	pfree(wholeFooter->data);
+	pfree(wholeFooter);
+
+	pfree(tableFooterMetadata->data);
+	pfree(tableFooterMetadata);
 }
```
anarazel: This many pfree()s usually is an indication that a memory context should be used instead.

cstore_fdw--1.5.sql
```diff
@@ -58,3 +60,37 @@ CREATE EVENT TRIGGER cstore_drop_event
     ON SQL_DROP
     EXECUTE PROCEDURE cstore_drop_trigger();
 
+CREATE TABLE public.pg_cstore_tables (logicalrelid regclass PRIMARY KEY);
```
pykello: High level suggestion:
 (1) Add data_size and block_row_count to pg_cstore_tables,
 (2) StripeHeader = StripeFooter + StripeMetadata (with unnecessary fields removed). Put this at beginning of stripe. No need to store StripeMetadata at StripeFooter.
 (3) With 1 and 2, we don't need TableFooter anymore.

For reads, we know the start position of next stripe after reading a stripe's header. We also know how much data is visible before current scan started. I guess these are sufficient to scan through the whole table.

For writes, by looking at data_size we know where to start appending.

I think this simplifies the layout, and removes the need for some of the functions that are not very easy to understand.

What do you think?

cstore_fdw.c
pykello: Can we use `DirectFunctionCall()` for plpgsql functions? (`registrer_cstore_table` is a plpgsql function)

Makefile
```diff
@@ -17,8 +17,8 @@ DATA = cstore_fdw--1.5.sql cstore_fdw--1.4--1.5.sql cstore_fdw--1.3--1.4.sql \
 REGRESS = create load query analyze data_types functions block_filtering drop \
 		  insert copyto alter truncate
 EXTRA_CLEAN = cstore.pb-c.h cstore.pb-c.c data/*.cstore data/*.cstore.footer \
-              sql/block_filtering.sql sql/create.sql sql/data_types.sql sql/load.sql \
-              sql/copyto.sql expected/block_filtering.out expected/create.out \
+              sql/block_filtering.sql sql/data_types.sql sql/load.sql \
+			  sql/copyto.sql expected/block_filtering.out \
               expected/data_types.out expected/load.out expected/copyto.out
```
mtuncer: ok

cstore_fdw--1.5.sql
```diff
@@ -58,3 +60,37 @@ CREATE EVENT TRIGGER cstore_drop_event
...
+ALTER TABLE public.pg_cstore_tables SET SCHEMA pg_catalog;
+
+CREATE OR REPLACE FUNCTION public.register_cstore_table(table_name oid)
+	RETURNS boolean
+	LANGUAGE plpgsql
```
mtuncer: Made it strict.
But I don't see the the difference. This was never meant to be called with null value.

cstore_fdw--1.5.sql
```diff
@@ -58,3 +60,37 @@ CREATE EVENT TRIGGER cstore_drop_event
...
+	END IF;
+
+	return false;
+END;
+$rct$;
+
+CREATE OR REPLACE FUNCTION public.unregister_cstore_table(table_name oid)
+	RETURNS void
+	LANGUAGE plpgsql
+	AS $uct$
```
mtuncer: ok

cstore_fdw.c

cstore_fdw.c
mtuncer: If cstore_fdw extension is dropped, it goes down with pg_cstore_tables that contain relation ids for existing cstore tables. 

If drop extension command issued on the cstore_fdw extension first cstore relation ids are cached, process utility is called to execute drop extension command, if it is successfull, cached relation's files are marked for removal.

This had to be done here because drop trigger is not called when the extension is dropped.

mtuncer: on second thought, that line looks extra, removing it since foreach would perform NIL comparison.

cstore_reader.c
```diff
@@ -80,43 +81,32 @@ static void DeserializeBlockData(StripeBuffers *stripeBuffers, uint64 blockIndex
...
-	pfree(tableFooterFilename);
-
-	tableFile = AllocateFile(filename, PG_BINARY_R);
-	if (tableFile == NULL)
+	if (tableFooter == NULL)
 	{
-		ereport(ERROR, (errcode_for_file_access(),
-						errmsg("could not open file \"%s\" for reading: %m",
-							   filename)));
+		ereport(ERROR, (errmsg("Could not read table file.")));
```
pykello: `ERRCODE_DATA_CORRUPTED` or `ERRCODE_FDW_ERROR` for error code?

cstore_fdw.c

cstore_fdw.c
mtuncer: Table ```pg_cstore_tables``` is going to be dropped right after this call.

Does it make sense to keep a lock on the table ? Only upside could be preventing a parallel create foreign table command from registering a new table. I don't know if we should worry about concurrent execution of drop extension and create/drop foreign table commands.

cstore_fdw.c
mtuncer: It is called from 2 places drop extension and from drop trigger function. I have rely on relation id being correct there.

cstore_metadata_serialization.c
mtuncer: we could live without it.

cstore_metadata_serialization.c
mtuncer: ok

cstore_reader.c

mtuncer: may be it is time to enable citus_indent for cstore files.

cstore_reader.c
mtuncer: That was to prevent any modification during read. Seems like locking only headerBuffer should be enough.

cstore_reader.c
mtuncer: ok

cstore_reader.c
```diff
@@ -1287,76 +1354,78 @@ ColumnDefaultValue(TupleConstr *tupleConstraints, Form_pg_attribute attributeFor
...
+		ereport(ERROR, (errmsg("could not read from file")));
 	}
 
-	return fileSize;
-}
-
+	while (remainingSize > 0)
+	{
+		Buffer buffer = ReadBufferExtended(relation, DATA_FORKNUM, blockNumber,
+										   RBM_NORMAL, NULL);
```
pykello: PostgreSQL‌ uses ring replacement strategy. See postgres/src/backend/storage/buffer/README.

cstore_reader.c
```diff
@@ -1287,76 +1354,78 @@ ColumnDefaultValue(TupleConstr *tupleConstraints, Form_pg_attribute attributeFor
...
-						errmsg("could not seek in file: %m")));
-	}
+		UnlockReleaseBuffer(buffer);
 
-	freadResult = fread(resultBuffer->data, size, 1, file);
-	if (freadResult != 1)
-	{
-		ereport(ERROR, (errmsg("could not read enough data from file")));
-	}
+		if (remainingSize == 0)
```
pykello: No need for this check, because we are doing `while (remainingSize >‌ 0)`.

cstore_fdw.c
```diff
@@ -254,6 +240,53 @@ cstore_ddl_event_end_trigger(PG_FUNCTION_ARGS)
...
+	List *argumentNames = NIL;
+	bool expandVariadic = false;
+	bool expandDefaults = false;
+#if PG_VERSION_NUM >= 90400
+	bool missingOK = true;
+#endif
+
+	qualifiedFunctionName = quote_qualified_identifier(schemaName, functionName);
+	qualifiedFunctionNameList = stringToQualifiedNameList(qualifiedFunctionName);
+	functionList = FuncnameGetCandidates(qualifiedFunctionNameList, argumentCount,
```
pykello: Can we use `LookupFuncName()` here? It seems to do most of the checks we do here.

pykello:
Also formatting in general, and update error messages and error codes.

cstore_reader.c
mtuncer: removed the check

cstore_reader.c
mtuncer: it was variable declaration

cstore_writer.c
mtuncer: ok

cstore_writer.c
mtuncer: done. please check the new usage

cstore_writer.c
mtuncer: foreign data wrapper api does not support creating unlogged tables. Therefore I provided logging parameter as an option to table creation. ```RelationNeedsWAL()``` does not provide that checking functionality.

cstore_writer.c
```diff
@@ -318,96 +326,221 @@ CStoreWriteRow(TableWriteState *writeState, Datum *columnValues, bool *columnNul
...
+	PageHeader headerBufferPageHeader = NULL;
+	char *headerPageContent = NULL;
+	uint32 startingBlock = InvalidBlockNumber;
+	uint32 blockCount = InvalidBlockNumber;
+	uint32 newStartingBlock = InvalidBlockNumber;
+	StringInfo tableFooterMetadata = NULL;
+
+	originalBlockCount = RelationGetNumberOfBlocksInFork(relation, FOOTER_FORKNUM);
+
+	appendBinaryStringInfo(wholeFooter, (const char *) &dataLength, sizeof(int32));
```
mtuncer: that is to leave space for actually setting it. see
```c
memcpy(wholeFooter->data, (const char*) &dataLength, sizeof(int32));
```

cstore_fdw.c
mtuncer: we can

cstore_reader.c
mtuncer: Some code after the loop required this early bailout without changing variables. Now they are gone, this check can be dropped.

cstore_reader.c
mtuncer: almost calls to ReadBufferExtended passes NULL as strategy in the postgres source. 

We can do ```GetAccessStrategy(BAS_BULKREAD)``` for bulk read operation.

cstore_reader.c
mtuncer: ok

cstore_writer.c
```diff
@@ -924,71 +1056,97 @@ AppendStripeMetadata(TableFooter *tableFooter, StripeMetadata stripeMetadata)
...
+		pageOffset = (pageHeader->pd_lower) - SizeOfPageHeaderData;
+
+		memcpy(pageData + pageOffset, (char *) data + dataOffset, copySize);
+
+		dataOffset += copySize;
+
+		pageHeader->pd_lower += copySize;
+		if (pageHeader->pd_lower >= pageHeader->pd_upper)
+		{
+			pageHeader->pd_flags |= PD_PAGE_FULL;
```
mtuncer: ok

cstore_writer.c
```diff
@@ -924,71 +1056,97 @@ AppendStripeMetadata(TableFooter *tableFooter, StripeMetadata stripeMetadata)
...
+		}
+
+		pageOffset = (pageHeader->pd_lower) - SizeOfPageHeaderData;
+
+		memcpy(pageData + pageOffset, (char *) data + dataOffset, copySize);
+
+		dataOffset += copySize;
+
+		pageHeader->pd_lower += copySize;
+		if (pageHeader->pd_lower >= pageHeader->pd_upper)
```
mtuncer: no it can't.

cstore_writer.c
```diff
@@ -172,17 +153,19 @@ CStoreBeginWrite(const char *filename, CompressionType compressionType,
 											   ALLOCSET_DEFAULT_INITSIZE,
 											   ALLOCSET_DEFAULT_MAXSIZE);
 
+	/* make sure file for data storage exists */
```
mtuncer: ok

cstore_writer.c
```diff
@@ -924,71 +1056,97 @@ AppendStripeMetadata(TableFooter *tableFooter, StripeMetadata stripeMetadata)
...
+		}
 
-	errorResult = ferror(file);
-	if (errorResult != 0)
-	{
-		ereport(ERROR, (errcode_for_file_access(),
-						errmsg("error in file: %m")));
-	}
-}
+		buffer = ReadBufferExtended(relation, DATA_FORKNUM, blockNumber, RBM_NORMAL, NULL);
```
mtuncer: Used strategy = BAS_BULKWRITE

cstore_fdw--1.5.sql
mtuncer: Shall we do those as a separate work on top of this one ?

cstore_fdw.c
```diff
@@ -299,8 +307,9 @@ CStoreProcessUtility(Node * parseTree, const char *queryString,
 
 		if (dropStmt->removeType == OBJECT_EXTENSION)
```
pykello: `DROP EXTENSION` should also fire our `cstore_drop_trigger()` in cstore_fdw--1.5.sql. Why not implement all these in there? (See the matrix at https://www.postgresql.org/docs/current/static/event-trigger-matrix.html)

cstore_fdw.c
pykello: @anarazel I am trying to understand your concern here. Considering this function in isolation, why shouldn't we drop the lock that was acquired in this function? Can you please elaborate more on the problems that dropping the lock might cause?

cstore_fdw.c
```diff
@@ -733,46 +719,58 @@ CStoreProcessAlterTableCommand(AlterTableStmt *alterStatement)
...
+	return cstoreRelationIdList;
+}
+
+
+ /* RemoveRelationStorage marks storage nodes for a relation for deletion */
+static void
+RemoveRelationStorage(Oid relationId)
+{
+	RelFileNode relFileNode = {MyDatabaseTableSpace, MyDatabaseId, relationId};
+	Relation relation = palloc0(sizeof(RelationData));
```
pykello: Why not use `relation_open()`?

cstore_fdw.c
```diff
@@ -1740,22 +1560,21 @@ TupleCountEstimate(RelOptInfo *baserel, const char *filename)
...
+
+	pageCount = footerBlockCount + dataBlockCount;
 
-	pageCount = (statBuffer.st_size + (BLCKSZ - 1)) / BLCKSZ;
-	if (pageCount < 1)
+	/* if the relation is empty, use default estimate for its size */
+	if (pageCount == 0)
 	{
-		pageCount = 1;
+		pageCount = 10;
```
pykello: why 10?

cstore_fdw.c
```diff
@@ -1304,22 +1151,7 @@ Datum
 cstore_clean_table_resources(PG_FUNCTION_ARGS)
```
pykello: Update comments.

cstore_fdw--1.5.sql
pykello: Ok. makes sense. Assuming you agree with what I‌ suggested, I am not reviewing table footer related changes. Also I‌ am not reviewing style or comments and am keeping that as last step after we make the high level changes and test them.

cstore_fdw.c
mtuncer: it does not. As soon as drop extension is called upon cstore_fdw, all triggers and functions are gone. Even though drop event might be fired, our trigger is not called.

cstore_fdw.c
mtuncer: When this was called from drop trigger, relation_open() fails because the relation was already dropped. This is to create a  **fake** relation which only has rd_node and rd_backend fields that are sufficient for storage removal.

cstore_fdw.c
```diff
@@ -254,6 +241,27 @@ cstore_ddl_event_end_trigger(PG_FUNCTION_ARGS)
...
+ * given relation id in pg_cstore_tables.
+ */
+static void
+RegisterCStoreTable(Oid relationId)
+{
+	char *schemaName = "public";
+	char *functionName = "register_cstore_table";
+	int argumentCount = 1;
+	const Oid argumentTypes[] = {OIDOID};
+	bool errorOK = false;
```
mtuncer: makes sense

cstore_fdw.c
mtuncer: done

cstore_fdw.c
mtuncer: see removed code
```c
if (statResult < 0)
{
	pageCount = footerBlockCount + dataBlockCount;
	statBuffer.st_size = 10 * BLCKSZ;		
}
```

cstore_writer.c
mtuncer: caller had a memory context available. Removed all these pfree's and let the caller handle memory context.

mtuncer:
@pykello @anarazel I am done with current feedback.  You can make another pass.

cstore_writer.c
```diff
@@ -310,104 +322,216 @@ CStoreWriteRow(TableWriteState *writeState, Datum *columnValues, bool *columnNul
...
+			newStartingBlock = 1;
+		}
+		else
+		{
+			newStartingBlock = startingBlock + blockCount;
+		}
+	}
+
+	dataOffset = 0;
+	memcpy(wholeFooter->data, (const char*) &dataLength, sizeof(int32));
```
pykello: Move up to right after we construct the footer data?

cstore_writer.c
```diff
@@ -310,104 +322,216 @@ CStoreWriteRow(TableWriteState *writeState, Datum *columnValues, bool *columnNul
...
+	BlockNumber actualBlockCount = 0;
+	BlockNumber currentBlockNumber = 0;
+	StringInfo wholeFooter = makeStringInfo();
+	int32 dataLength = 0;
+	int32 dataOffset = 0;
+	const int32 blockDataSize =  CSTORE_PAGE_DATA_SIZE;
+	BlockNumber headerBlockNumber = InvalidBlockNumber;
+	Buffer headerBuffer = InvalidBuffer;
+	Page headerPage = NULL;
+	PageHeader headerBufferPageHeader = NULL;
```
pykello: header* to metadata* or footerMetadata* to not confuse with buffer headers?

cstore_writer.c
```diff
@@ -310,104 +322,216 @@ CStoreWriteRow(TableWriteState *writeState, Datum *columnValues, bool *columnNul
...
-
-	tableFooterFile = AllocateFile(tableFooterFilename->data, PG_BINARY_W);
-	if (tableFooterFile == NULL)
-	{
-		ereport(ERROR, (errcode_for_file_access(),
-						errmsg("could not open file \"%s\" for writing: %m",
-							   tableFooterFilename->data)));
-	}
-
+	BlockNumber originalBlockCount = 0;
```
pykello: Naming suggestion: `forkBlockCount`?

cstore_writer.c
```diff
@@ -310,104 +322,216 @@ CStoreWriteRow(TableWriteState *writeState, Datum *columnValues, bool *columnNul
...
+	BlockNumber actualBlockCount = 0;
```
pykello: Naming suggestion: `footerDataBlockCount`?

cstore_writer.c
```diff
@@ -310,104 +322,216 @@ CStoreWriteRow(TableWriteState *writeState, Datum *columnValues, bool *columnNul
...
+	BlockNumber currentBlockNumber = 0;
```
pykello: Naming suggestion: `blockIndex`?

cstore_writer.c
```diff
@@ -310,104 +322,216 @@ CStoreWriteRow(TableWriteState *writeState, Datum *columnValues, bool *columnNul
...
+	int32 dataLength = 0;
+	int32 dataOffset = 0;
+	const int32 blockDataSize =  CSTORE_PAGE_DATA_SIZE;
+	BlockNumber headerBlockNumber = InvalidBlockNumber;
+	Buffer headerBuffer = InvalidBuffer;
+	Page headerPage = NULL;
+	PageHeader headerBufferPageHeader = NULL;
+	char *headerPageContent = NULL;
+	uint32 startingBlock = InvalidBlockNumber;
+	uint32 blockCount = InvalidBlockNumber;
```
pykello: Naming suggestion: `previousFooterDataBlockCount`?

cstore_writer.c
```diff
@@ -310,104 +322,216 @@ CStoreWriteRow(TableWriteState *writeState, Datum *columnValues, bool *columnNul
...
+	uint32 newStartingBlock = InvalidBlockNumber;
```
pykello: Naming suggestion:
- `startingBlock` -> `previousBlockOffset` or `previousStartingBlock`
- `newStartingBlock` -> `blockOffset` or `startingBlock`

Or if you want to use `new.*`, update all related variable names to reflect the relationship. I am not against `new.*`, just make variable naming uniform and more descriptive.

cstore_metadata_serialization.c
pykello: Update as Andres suggested?

cstore_fdw--1.5.sql
```diff
@@ -58,3 +60,37 @@ CREATE EVENT TRIGGER cstore_drop_event
...
 
+CREATE TABLE public.pg_cstore_tables (logicalrelid regclass PRIMARY KEY);
+ALTER TABLE public.pg_cstore_tables SET SCHEMA pg_catalog;
+
+CREATE OR REPLACE FUNCTION public.register_cstore_table(table_name oid)
+	RETURNS boolean
+	LANGUAGE plpgsql STRICT
+	AS $rct$
+BEGIN
+    IF EXISTS(
```
mtuncer: exposing a UDF for a simple select query looks like an overkill.

Moving implementation of register/unregister functions inside c code looks better to me.

cstore_fdw--1.5.sql
```diff
@@ -58,3 +60,37 @@ CREATE EVENT TRIGGER cstore_drop_event
...
+		SELECT fdwname FROM pg_foreign_table ft, pg_foreign_server fs, pg_foreign_data_wrapper fdw
+			WHERE 
+				ft.ftrelid = table_name AND
+				fs.oid = ft.ftserver AND
+				fdw.oid = fs.srvfdw AND
+				fdw.fdwname = 'cstore_fdw'
+				) AND
+		NOT EXISTS(SELECT * FROM pg_cstore_tables WHERE logicalrelid = table_name)
```
mtuncer: moving UDF to c code would resolve this.

cstore_fdw.c
```diff
@@ -1034,157 +1010,30 @@ DistributedWorkerCopy(CopyStmt *copyStatement)
...
-	if (dataFileStatResult != 0)
-	{
-		ereport(ERROR, (errcode_for_file_access(),
-						errmsg("could not stat file \"%s\": %m", dataFilename)));
-	}
-
-	footerFilename = makeStringInfo();
-	appendStringInfo(footerFilename, "%s%s", dataFilename,
-					 CSTORE_FOOTER_FILE_SUFFIX);
+	relation = heap_open(relationId, AccessExclusiveLock);
```
mtuncer: correct, already encountered an issue during cloud testing. fixed.

cstore_fdw.c
```diff
@@ -1506,29 +1336,18 @@ ValidateForeignTableOptions(char *filename, char *compressionTypeString,
...
-					 databaseOid, relationFileOid);
-
-	relation_close(relation, AccessShareLock);
-
-	return cstoreFilePath->data;
+	/* check if the provided logging flag is valid */
+	if (loggingString != NULL)
+	{
+		if (strncmp(LOGGING_STRING_TRUE, loggingString, NAMEDATALEN) != 0 &&
+			strncmp(LOGGING_STRING_FALSE, loggingString, NAMEDATALEN) != 0)
```
mtuncer: parse_pool accepts variety of input values. I am trying to restrict this option specifically to "true" or "false"

cstore_fdw.c
```diff
@@ -1872,22 +1690,17 @@ ColumnList(RelOptInfo *baserel, Oid foreignTableId)
...
-	CStoreFdwOptions *cstoreFdwOptions = CStoreGetOptions(foreignTableId);
-
-	ExplainPropertyText("CStore File", cstoreFdwOptions->filename, explainState);
-
 	/* supress file size if we're not showing cost details */
 	if (explainState->costs)
 	{
-		struct stat statBuffer;
+		Oid foreignTableId = RelationGetRelid(scanState->ss.ss_currentRelation);
+		Relation relation = heap_open(foreignTableId, AccessExclusiveLock);
```
mtuncer: AccessShareLock would do fine. fixed

cstore_fdw.c
```diff
@@ -1400,18 +1234,16 @@ CStoreGetOptions(Oid foreignTableId)
 	{
 		blockRowCount = pg_atoi(blockRowCountString, sizeof(int32), 0);
 	}
-
-	/* set default filename if it is not provided */
-	if (filename == NULL)
+	if (loggingString != NULL)
 	{
-		filename = CStoreDefaultFilePath(foreignTableId);
+		logging = strncmp(loggingString, LOGGING_STRING_TRUE, NAMEDATALEN) == 0;
```
mtuncer: we could, but it can only have true or false values. parse_bool is more general purpose

cstore_fdw.h
```diff
@@ -272,21 +282,23 @@ typedef struct TableReadState
...
 	TableFooter *tableFooter;
-	StringInfo tableFooterFilename;
 	CompressionType compressionType;
+	bool logging;
 	TupleDesc tupleDescriptor;
 	FmgrInfo **comparisonFunctionArray;
 	uint64 currentFileOffset;
 	Relation relation;
+	BlockNumber activeBlockNumber;
+
```
mtuncer: I will have to run citus_indent on all files. It is generating far more changes than we want to review in the context of this pr.

cstore_metadata_serialization.c
```diff
@@ -579,3 +580,78 @@ ProtobufBinaryToDatum(ProtobufCBinaryData protobufBinary, bool datumTypeByValue,
 
 	return datum;
 }
+
+
+/* DeserializeFooterMetadata deserializes footer metadata to extract where footer
```
mtuncer: fixed

cstore_metadata_serialization.c
```diff
@@ -579,3 +580,78 @@ ProtobufBinaryToDatum(ProtobufCBinaryData protobufBinary, bool datumTypeByValue,
...
+	memcpy(startingPage, data + offset, sizeof(uint32));
+	offset += sizeof(uint32);
+
+	memcpy(pageCount, data + offset, sizeof(uint32));
+
+	return true;
+}
+
+
+/* SerializeTableFooterMetadata serializes footer metadata. Serialized string
```
mtuncer: fixed

cstore_reader.c
```diff
@@ -147,78 +137,168 @@ CStoreBeginRead(const char *filename, TupleDesc tupleDescriptor,
...
+	 * DeserializeTableFooter expects footer data without postscript, reduce len
+	 * field to trim the right end of footer data.
+	 */
+	footerData->len = footerLength;
+	tableFooter = DeserializeTableFooter(footerData);
+
+	pfree(footerData->data);
+	pfree(footerData);
+
+
```
mtuncer: fixed.

indent will be run on whole branch separately

cstore_fdw--1.5.sql
mtuncer: For the time being, I am not dropping support for 9.4 and earlier. I am going to open an issue on moving aforementioned UDFs to c code.

cstore_reader.c
```diff
@@ -147,78 +137,168 @@ CStoreBeginRead(const char *filename, TupleDesc tupleDescriptor,
...
 	readState->deserializedBlockIndex = -1;
+	readState->relation = relation;
 
 	return readState;
 }
 
 
 /*
- * CStoreReadFooter reads the cstore file footer from the given file. First, the
+ * CStoreReadFooter reads the cstore file footer for the given relation. First, the
```
mtuncer: fixed

cstore_reader.c
```diff
@@ -147,78 +137,168 @@ CStoreBeginRead(const char *filename, TupleDesc tupleDescriptor,
...
-	tableFooterFile = AllocateFile(tableFooterFilename->data, PG_BINARY_R);
-	if (tableFooterFile == NULL)
+	footerData = ReadFooterData(relation);
+	if (footerData == NULL)
 	{
-		ereport(ERROR, (errcode_for_file_access(),
-						errmsg("could not open file \"%s\" for reading: %m",
-							   tableFooterFilename->data),
-						errhint("Try copying in data to the table.")));
+		return NULL;
```
mtuncer: CStoreBeginWrite also calls this function now, therefore it is okay to return null. Callers are handling NULL values themselves.

cstore_reader.c
```diff
@@ -147,78 +137,168 @@ CStoreBeginRead(const char *filename, TupleDesc tupleDescriptor,
...
+	uint32 blockCount = 0;
+
+	totalBlockCount = RelationGetNumberOfBlocksInFork(relation, FOOTER_FORKNUM);
+
+	if (totalBlockCount == 0)
+	{
+		return NULL;
+	}
+
+	headerBuffer =  ReadBufferExtended(relation, FOOTER_FORKNUM, 0, RBM_NORMAL,
```
mtuncer: ok

cstore_reader.c
```diff
@@ -147,78 +137,168 @@ CStoreBeginRead(const char *filename, TupleDesc tupleDescriptor,
...
+									   NULL);
+	LockBuffer(headerBuffer, BUFFER_LOCK_SHARE);
+	headerPage = BufferGetPage(headerBuffer);
```

cstore_reader.c
```diff
@@ -147,78 +137,168 @@ CStoreBeginRead(const char *filename, TupleDesc tupleDescriptor,
...
+	headerData = PageGetContents(headerPage);
```

cstore_reader.c
```diff
@@ -147,78 +137,168 @@ CStoreBeginRead(const char *filename, TupleDesc tupleDescriptor,
...
-	return tableFooter;
+	firstBuffer = ReadBufferExtended(relation, FOOTER_FORKNUM, startingBlock,
+									 RBM_NORMAL, NULL);
+
+	LockBuffer(firstBuffer, BUFFER_LOCK_SHARE);
+	page = BufferGetPage(firstBuffer);
+	pageHeader = (PageHeader) page;
+	pageData = PageGetContents(page);
+
+	memcpy(&fileLength, pageData, headerLength);
```
mtuncer: after some thought, length of table footer does not do much help. We already store useful data payload length in Page headers.

We should consider removing that 4 bytes from beginning of table footer.

cstore_reader.c
```diff
@@ -147,78 +137,168 @@ CStoreBeginRead(const char *filename, TupleDesc tupleDescriptor,
...
+
+	if (fileLength < CSTORE_POSTSCRIPT_SIZE_LENGTH)
+	{
+		ereport(ERROR, (errmsg("invalid cstore file")));
```
mtuncer: This indicates something is corrupt.
anyway removing this part from footer data looks like move viable

cstore_reader.c
```diff
@@ -397,37 +477,25 @@ FreeColumnBlockDataArray(ColumnBlockData **blockDataArray, uint32 columnCount)
...
-	pfree(tableFooterFilename->data);
-	pfree(tableFooterFilename);
-
-	tableFile = AllocateFile(filename, PG_BINARY_R);
-	if (tableFile == NULL)
+	if (tableFooter == NULL)
 	{
-		ereport(ERROR, (errcode_for_file_access(),
-						errmsg("could not open file \"%s\" for reading: %m", filename)));
+		ereport(ERROR, (ERRCODE_FDW_ERROR, errmsg("Could not read table file.")));
```
mtuncer: does not make much difference but ok.

cstore_reader.c
```diff
@@ -1291,76 +1361,70 @@ ColumnDefaultValue(TupleConstr *tupleConstraints, Form_pg_attribute attributeFor
 }
 
 
-/* Returns the size of the given file handle. */
-static int64
-FileSize(FILE *file)
+/* ReadFromFile reads the given segment from the relation data file. */
+static StringInfo
+ReadFromFile(Relation relation, uint64 offset, uint32 size)
```
mtuncer: They are files actually, just that we can only read in blocks.

cstore_reader.c
```diff
@@ -1291,76 +1361,70 @@ ColumnDefaultValue(TupleConstr *tupleConstraints, Form_pg_attribute attributeFor
...
+	StringInfo resultBuffer = NULL;
+	BufferAccessStrategy strategy = GetAccessStrategy(BAS_BULKREAD);
 
-	fileSize = ftello(file);
-	if (fileSize == -1)
+	if (size == 0)
 	{
-		ereport(ERROR, (errcode_for_file_access(),
-						errmsg("could not get position in file: %m")));
+		return NULL;
```
mtuncer: It is not an error. Caller requested 0 size read.

cstore_reader.c
```diff
@@ -1291,76 +1361,70 @@ ColumnDefaultValue(TupleConstr *tupleConstraints, Form_pg_attribute attributeFor
...
-	int freadResult = 0;
-	int fileError = 0;
-
-	StringInfo resultBuffer = makeStringInfo();
+	resultBuffer = makeStringInfo();
 	enlargeStringInfo(resultBuffer, size);
-	resultBuffer->len = size;
 
-	if (size == 0)
+	blockCount = RelationGetNumberOfBlocksInFork(relation,DATA_FORKNUM);
```
mtuncer: ok

cstore_writer.c
```diff
@@ -19,18 +19,29 @@
...
 #include "optimizer/var.h"
 #include "port.h"
-#include "storage/fd.h"
+#include "storage/smgr.h"
 #include "utils/memutils.h"
 #include "utils/lsyscache.h"
 #include "utils/rel.h"
 
+#if PG_VERSION_NUM >= 90400
+#define LOG_NEWPAGE_BUFFER(buffer) log_newpage_buffer(buffer, false)
```
mtuncer: fixed

cstore_writer.c
```diff
@@ -191,12 +173,42 @@ CStoreBeginWrite(const char *filename, CompressionType compressionType,
...
 	return writeState;
 }
 
 
 /*
+ * EnsureDataForkExists checks if data fork exists for writing data, and
+ * creates if it is not present. Relation must be previously opened with
+ * exclusive lock (AccessExclusive or ShareUpdateExclusive).
+ */
+static void EnsureDataForkExists(Relation relation, bool logging)
```
mtuncer: done

cstore_writer.c
```diff
@@ -310,104 +322,216 @@ CStoreWriteRow(TableWriteState *writeState, Datum *columnValues, bool *columnNul
...
+ * postscript size as the last byte of the footer data.
+ * After preparing the footer data the function reads the current footer
+ * metadata to decide where to write to make sure that the current footer data
+ * is not overwritten. It writes the footer data to correct place and finally
+ * updates footer metadata about where footer data is stored. Reclamation of
+ * created StringInfo buffers are performed in caller function's memory context.
  */
 static void
-CStoreWriteFooter(StringInfo tableFooterFilename, TableFooter *tableFooter)
+CStoreWriteFooter(TableFooter *tableFooter, Relation relation, bool loggingEnabled)
```
mtuncer: Some logic from WriteToFile could be used here. I will revisit this.

cstore_writer.c
mtuncer: ok

cstore_writer.c
mtuncer: ok

cstore_writer.c
mtuncer: ok

cstore_writer.c
mtuncer: removed new notation

cstore_writer.c
```diff
@@ -310,104 +322,216 @@ CStoreWriteRow(TableWriteState *writeState, Datum *columnValues, bool *columnNul
...
+	Buffer headerBuffer = InvalidBuffer;
+	Page headerPage = NULL;
+	PageHeader headerBufferPageHeader = NULL;
+	char *headerPageContent = NULL;
+	uint32 startingBlock = InvalidBlockNumber;
+	uint32 blockCount = InvalidBlockNumber;
+	uint32 newStartingBlock = InvalidBlockNumber;
+	StringInfo tableFooterMetadata = NULL;
+
+	originalBlockCount = RelationGetNumberOfBlocksInFork(relation, FOOTER_FORKNUM);
```
mtuncer: ok

cstore_writer.c
mtuncer: ok

cstore_writer.c
```diff
@@ -924,71 +1047,99 @@ AppendStripeMetadata(TableFooter *tableFooter, StripeMetadata stripeMetadata)
...
 
-/* Writes the given data to the given file pointer and checks for errors. */
+/*
+ * WriteToFile appends provided data to the active page. If data does
+ * not fit into remaining space in page, a new page is allocated until whole
+ * data is written.
+ */
 static void
-WriteToFile(FILE *file, void *data, uint32 dataLength)
+WriteToFile(TableWriteState *writeState, void *data, uint32 dataLength)
```
mtuncer: needs refactoring. I would like to make a function something like WriteToFork and call this function to write both data and footer

mtuncer:
@pykello can you make another pass

missing things are
- writetofile/writefooter function needs refactoring
- cstore version needs to be upgraded and merge conflict should be resolved
- register/unregister cstore table UDFs needs to be ported to c. (this is will be done later)

cstore_writer.c
```diff
@@ -310,104 +323,228 @@ CStoreWriteRow(TableWriteState *writeState, Datum *columnValues, bool *columnNul
...
+	StringInfo wholeFooter = makeStringInfo();
+	int32 dataLength = 0;
+	int32 dataOffset = 0;
+	const int32 blockDataSize =  CSTORE_PAGE_DATA_SIZE;
+	BlockNumber footerHeaderBlockNumber = InvalidBlockNumber;
+	Buffer footerHeaderBuffer = InvalidBuffer;
+	Page footerHeaderPage = NULL;
+	PageHeader footerHeaderBufferPageHeader = NULL;
+	char *footerHeaderPageContent = NULL;
+	uint32 existingFoooterStartingBlock = InvalidBlockNumber;
```
pykello: Foooter -> Footer

cstore_metadata_serialization.c
```diff
@@ -579,3 +580,80 @@ ProtobufBinaryToDatum(ProtobufCBinaryData protobufBinary, bool datumTypeByValue,
...
+	const uint32 footerMetadataLength = magicLength + 4 * sizeof(uint32);
+	uint32 majorVersion = 0;
+	uint32 minorVersion = 0;
+
+	*startingPage = InvalidBuffer;
+	*pageCount = InvalidBuffer;
+
+	if (length < footerMetadataLength)
+	{
+		return false;
```
anarazel: Why don't we throw errors here?

cstore_reader.c
```diff
@@ -147,78 +137,167 @@ CStoreBeginRead(const char *filename, TupleDesc tupleDescriptor,
...
+	}
+
+	footerData = makeStringInfo();
+
+	pageDataLength = pageHeader->pd_lower - SizeOfPageHeaderData - sizeofFooterLengthType;
+
+	appendBinaryStringInfo(footerData, pageData + sizeofFooterLengthType, pageDataLength);
+
+	UnlockReleaseBuffer(firstBuffer);
+
```
anarazel: Might be worth to do a `enlargeStringInfo()` before iterating, to avoid repeated reallocations.

cstore_writer.c
```diff
@@ -161,6 +140,8 @@ CStoreBeginWrite(const char *filename, CompressionType compressionType,
 		comparisonFunctionArray[columnIndex] = comparisonFunction;
 	}
 
+
+
```
anarazel: three newlines? ;)

cstore_writer.c
```diff
@@ -191,12 +173,43 @@ CStoreBeginWrite(const char *filename, CompressionType compressionType,
...
+		if (logging)
+		{
+			log_smgrcreate(&srel->smgr_rnode.node, DATA_FORKNUM);
+		}
+	}
+
+	Assert(smgrexists(srel, DATA_FORKNUM));
+
+	RelationCloseSmgr(relation);
```
anarazel: Why are you closing here?

cstore_writer.c
```diff
@@ -310,104 +323,228 @@ CStoreWriteRow(TableWriteState *writeState, Datum *columnValues, bool *columnNul
...
+	PageInit(footerHeaderPage, BLCKSZ, 0);
+	memcpy(footerHeaderPageContent, tableFooterMetadata->data, tableFooterMetadata->len);
+	footerHeaderBufferPageHeader->pd_lower = SizeOfPageHeaderData + tableFooterMetadata->len;
+	MarkBufferDirty(footerHeaderBuffer);
+
+	/*
+	 * Changes to headerBuffer is logged regardless of logging setting to enable
+	 * recovery after crash.
+	 */
+	LOG_NEWPAGE_BUFFER(footerHeaderBuffer);
```
anarazel: Why isn't this guarded by `if (loggingEnabled)`?

Makefile
```diff
@@ -17,9 +17,9 @@ DATA = cstore_fdw--1.5.sql cstore_fdw--1.4--1.5.sql cstore_fdw--1.3--1.4.sql \
 REGRESS = create load query analyze data_types functions block_filtering drop \
 		  insert copyto alter truncate
 EXTRA_CLEAN = cstore.pb-c.h cstore.pb-c.c data/*.cstore data/*.cstore.footer \
-              sql/block_filtering.sql sql/create.sql sql/data_types.sql sql/load.sql \
-              sql/copyto.sql expected/block_filtering.out expected/create.out \
-              expected/data_types.out expected/load.out expected/copyto.out
+			  sql/block_filtering.sql sql/data_types.sql sql/load.sql \
+			  sql/copyto.sql expected/block_filtering.out \
+			  expected/data_types.out expected/load.out expected/copyto.out
```
mtuncer: these sql files are generated out of source files just like some out files in expected folder.
indentation should be better this way

Makefile
```diff
@@ -33,6 +33,7 @@ endif
 #
 
 PG_CONFIG = pg_config
+
```
mtuncer: ok

cstore_fdw--1.5.sql
```diff
@@ -58,3 +60,37 @@ CREATE EVENT TRIGGER cstore_drop_event
...
+	THEN
+		INSERT INTO pg_cstore_tables SELECT table_name;
+		RETURN true;
+	END IF;
+
+	RETURN false;
```
mtuncer: to prevent multiple insertions. I guess declaring logical relid as primary key would do that for free.

mtuncer: removing the line

cstore_fdw.c
```diff
@@ -733,46 +722,58 @@ CStoreProcessAlterTableCommand(AlterTableStmt *alterStatement)
...
+
+	return cstoreRelationIdList;
+}
+
+
+/* RemoveRelationStorage marks storage nodes for a relation for deletion */
+static void
+RemoveRelationStorage(Oid relationId)
+{
+	RelFileNode relFileNode = { MyDatabaseTableSpace, MyDatabaseId, relationId };
```
mtuncer: done

cstore_metadata_serialization.c
mtuncer: done

anarazel:
On 2017-07-28 13:44:32 -0700, Murat Tuncer wrote:
> to prevent multiple insertions. I guess declaring logical relid as primary key would do that for free.

Right. Why would it ever happen? Also, using NOT EXISTS() isn't actually
concurrency safe (consider two transactions inserting that both haven't
committed yet).

cstore_reader.c
mtuncer: ok

cstore_writer.c
mtuncer: it is new :), fixed

cstore_writer.c
mtuncer: we won't be needing it until flush stripe.  

are we supposed to keep it open ?

cstore_writer.c
mtuncer: I originally thought it would be nicer for crash recovery to log changes to metadata regardless of logging setting

cstore_writer.c
anarazel: There's no reason not to, and it sometimes isn't entirely cheap to reopen.

Makefile
anarazel: Even in that case the lines aren't indented uniformly (mix tab vs space)

cstore_writer.c
anarazel: Hm. Seems like we really should trigger tables to be truncated after an unclean shutdown?

mtuncer:
bumped version to 2.0 and prevented upgrade from 1.x

cstore_writer.c
mtuncer: ok, removed call to close

cstore_fdw.c
```diff
@@ -1872,22 +1690,17 @@ ColumnList(RelOptInfo *baserel, Oid foreignTableId)
...
+		BlockNumber pageCount = PageCount(relation);
+		long fileSize = pageCount * BLCKSZ;
 
-		int statResult = stat(cstoreFdwOptions->filename, &statBuffer);
-		if (statResult == 0)
-		{
-			ExplainPropertyLong("CStore File Size", (long) statBuffer.st_size,
-								explainState);
-		}
+		heap_close(relation, AccessShareLock);
```
mtuncer: done

cstore_fdw.c
```diff
@@ -1019,172 +995,45 @@ DistributedTable(Oid relationId)
...
-	footerFileStatResult = stat(footerFilename->data, &footerFileStatBuffer);
-	if (footerFileStatResult != 0)
-	{
-		ereport(ERROR, (errcode_for_file_access(),
-						errmsg("could not stat file \"%s\": %m",
-								footerFilename->data)));
-	}
+	relation = heap_open(relationId, AccessShareLock);
+	pageCount = PageCount(relation);
+	heap_close(relation, AccessShareLock);
```
mtuncer: done

cstore_fdw.c
```diff
@@ -733,46 +722,58 @@ CStoreProcessAlterTableCommand(AlterTableStmt *alterStatement)
...
-				/*
-				 * Skip files that are placed in default location, they are handled
-				 * by sql drop trigger. Both paths are generated by code, use
-				 * of strcmp is safe here.
-				 */
-				if (strcmp(defaultfilename, cstoreFdwOptions->filename) == 0)
-				{
-					continue;
-				}
+	scanDesc = heap_beginscan(heapRelation, SnapshotSelf, 0, NULL);
```
mtuncer: due to 9.3 used GetTransactionSnapshot() instead

cstore_fdw--1.5.sql
mtuncer: I removed both register/unregister UDFs. Now they are static functions. There is no need to expose them anymore.

mtuncer:
@anarazel removed 2 new UDFs (register/unregister) and implemented the logic inside cstore_fdw.c

mtuncer:
@pykello @anarazel 

I made 3 more changes since your last look
- register/unregister UDFs are  removed and implemented inside c code. There is no need to expose them anymore
- new field is added to pg_cstore_tables tables to keep track of actual rel file node number since it can be changed during truncate
- file forks used by footer and data files are switched to make pg_relation_size return more accurate values

cstore_fdw.c
```diff
@@ -254,6 +248,139 @@ cstore_ddl_event_end_trigger(PG_FUNCTION_ARGS)
...
+{
+	Oid pgCstoreTablesOid = InvalidOid;
+	Relation pgCstoreTables = NULL;
+	TupleDesc tupleDescriptor = NULL;
+	HeapTuple heapTuple = NULL;
+	Datum values[] = { ObjectIdGetDatum(relationId), ObjectIdGetDatum(fileNodeOid) };
+	bool isNulls[] = { false, false };
+
+	pgCstoreTablesOid = get_relname_relid(CSTORE_TABLES_NAME, PG_CATALOG_NAMESPACE);
+
```
pykello: Remove empty line?

cstore_fdw.c
```diff
@@ -254,6 +248,139 @@ cstore_ddl_event_end_trigger(PG_FUNCTION_ARGS)
...
+{
+	Oid pgCstoreTablesOid = InvalidOid;
+	Relation pgCstoreTables = NULL;
+	HeapTuple heapTuple = NULL;
+	int scanKeyCount = 1;
+	ScanKeyData scanKey[1];
+	HeapScanDesc scanDescriptor = NULL;
+
+	pgCstoreTablesOid = get_relname_relid(CSTORE_TABLES_NAME, PG_CATALOG_NAMESPACE);
+
```
pykello: Remove empty line? (What is the style guide for empty lines? What I remember is that when we set a variable and then check for its value we didn't put empty line)

cstore_fdw.c
```diff
@@ -254,6 +248,139 @@ cstore_ddl_event_end_trigger(PG_FUNCTION_ARGS)
 
 
 /*
+ * RegisterCStoreTable registers provided cstore table relation id by
+ * adding a new row to pg_cstore_tables.
+ */
+static void
+RegisterCStoreTable(Oid relationId, Oid fileNodeOid)
```
pykello: Alternatively, we could use `SPI_Connect()`, `SPI_Execute()`, ... to implement this function and the unregister function if that makes things simpler.

mtuncer: using SPI would be at almost equally complexity.

cstore_fdw.c
mtuncer: correct. We are recently more liberal on using empty lines.
removed the line

cstore_fdw.c
mtuncer: done

cstore_fdw.c
```diff
@@ -733,46 +847,65 @@ CStoreProcessAlterTableCommand(AlterTableStmt *alterStatement)
...
+ * Postgres does not mark foreign tables' file nodes for deletion since
+ * it never creates it. Here we create a fake relation, associate this
+ * relation with cstore tables file node, then mark this relation for
+ * deletion so that file node is dropped when the transaction commits.
+ */
+static void
+RemoveRelationStorage(Oid relationId, Oid fileNodeOid)
+{
+	RelFileNode relFileNode = { MyDatabaseTableSpace, MyDatabaseId, fileNodeOid };
+	Relation relation = palloc0(sizeof(RelationData));
```
anarazel: Why aren't we just allocating this on the stack?

cstore_fdw.c
```diff
@@ -314,36 +442,37 @@ CStoreProcessUtility(Node * parseTree, const char *queryString,
 				Assert(IsA(object, List));
 				objectName = strVal(linitial((List *) object));
 #endif
-
+				/* record cstore relation ids if cstore extension is to be dropped */
 				if (strncmp(CSTORE_FDW_NAME, objectName, NAMEDATALEN) == 0)
 				{
-					removeCStoreDirectory = true;
+					cstoreTableList = CStoreTableList();
```
anarazel: I'd open an issue to replace this with a dependency from the relation to the extension, once other parts are ready. Or at least add an XXX comment.

cstore_fdw.c
```diff
@@ -314,36 +442,37 @@ CStoreProcessUtility(Node * parseTree, const char *queryString,
...
-								  destReceiver, completionTag);
 
-			foreach(fileListCell, droppedTables)
-			{
-				char *fileName = lfirst(fileListCell);
+			CALL_PREVIOUS_UTILITY(parseTree, queryString, context,
+					paramListInfo, destReceiver, completionTag);
 
-				DeleteCStoreTableFiles(fileName);
+			/* mark cached cstore relations' storage for deletion*/
```
anarazel: "cached" seems misleading.

cstore_fdw.c
```diff
@@ -733,46 +847,65 @@ CStoreProcessAlterTableCommand(AlterTableStmt *alterStatement)
...
-	if (dropStatement->removeType == OBJECT_FOREIGN_TABLE)
+	cstoreTablesOid = get_relname_relid(CSTORE_TABLES_NAME, PG_CATALOG_NAMESPACE);
+	if (cstoreTablesOid == InvalidOid)
 	{
-		ListCell *dropObjectCell = NULL;
-		foreach(dropObjectCell, dropStatement->objects)
-		{
-			List *tableNameList = (List *) lfirst(dropObjectCell);
-			RangeVar *rangeVar = makeRangeVarFromNameList(tableNameList);
+		/* the pg_cstore_tables table does not exist */
```
anarazel: When's that the case? I presume once the extension is dropped? Would be good to mention here.

cstore_fdw.c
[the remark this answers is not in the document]
anarazel: Why?

cstore_fdw.c
anarazel: That's not an explanation.

cstore_reader.c
```diff
@@ -1291,76 +1359,70 @@ ColumnDefaultValue(TupleConstr *tupleConstraints, Form_pg_attribute attributeFor
...
-	resultBuffer->len = size;
 
-	if (size == 0)
+	blockCount = RelationGetNumberOfBlocksInFork(relation, DATA_FORKNUM);
+	if (blockNumber > blockCount)
 	{
-		return resultBuffer;
+		ereport(ERROR, (errcode(ERRCODE_FDW_ERROR),
+						errmsg("could not read from file")));
 	}
```
anarazel: Hm, isn't this just a repeat of a check that postgres also does?

cstore_writer.c
```diff
@@ -310,104 +319,229 @@ CStoreWriteRow(TableWriteState *writeState, Datum *columnValues, bool *columnNul
...
+	MemoryContextSwitchTo(oldContext);
 
 	MemoryContextDelete(writeState->stripeWriteContext);
 	list_free_deep(writeState->tableFooter->stripeMetadataList);
 	pfree(writeState->tableFooter);
-	pfree(writeState->tableFooterFilename->data);
-	pfree(writeState->tableFooterFilename);
 	pfree(writeState->comparisonFunctionArray);
 	FreeColumnBlockDataArray(writeState->blockDataArray, columnCount);
 	pfree(writeState);
```
anarazel: Why isn't this just memory inside the stripeWriteContext?

cstore_fdw.c
mtuncer: This was a very minor detail. I just wanted to preserve the existing behavior.

Removed default page count and return the actual number instead.

cstore_writer.c
mtuncer: fixed

cstore_fdw.c
mtuncer: it really does not make much difference. 

Used a stack variable of type RelationData and used its reference.

cstore_fdw.c
mtuncer: I did not get this. 
There is already a dependency from cstore relations to cstore extension. CStore tables are dropped when the extension is dropped.

The issue here is keeping track of actual file nodes. If I can setup a dependency from a file node to the extension or even a relation it would be fine.  

This piece of code determines which file nodes should be dropped as a result of extension drop.

cstore_fdw.c
mtuncer: removed the word

cstore_fdw.c
mtuncer: It should never happen.
Replaced if with an Assert.

cstore_reader.c
mtuncer: This felt easier to debug. Postgres' check is buried deep down and difficult to follow for an untrained eye.

cstore_writer.c
mtuncer: stripeWriteContext is memory context per stripe. It gets reset upon each stripe flush.

writeState is per whole write operation.  It needs to survive stripeWriteContext resets.

mtuncer:
@anarazel can you make another pass ?

cstore_reader.c
anarazel: -1

cstore_writer.c
anarazel: Ok, there probably still should be a memory context for the whole write, rather than this granular manual management. I'm ok with opening a new issue for that tho.

cstore_fdw.c
anarazel: The confusion is/was that dependencies should normally prevent that from happening. Except I guess that if you drop the extension with CASCADE, you can drop both extension and tables at the same time. But even then the tables should be dropped first, no?  Needs explanatory comment.

cstore_reader.c
anarazel: But that means the code is unreachable, and you'll get a more descriptive error first. It's better to add a comment saying that it's "guaranteed" to succeed or never reach that part of the code.

mtuncer: replace this with asserd

cstore_writer.c
mtuncer: introduce a context per command and clear it when  done (in  a different PR)

mtuncer: logged this in #162

cstore_fdw.c
mtuncer: added comment

sumedhpathak:
🍾  Awesome!

PR #147, opened 2017-06-14
Read stripe skip list only if column is referenced by the query

### LINKED ISSUE #136: Deserializing StripeSkipList for columns which are not projected
Hello

We seem to deserialize StripeSkipList for columns which are not projected as well . I had a use case where I had 516 columns and only 16 columns had values. While projecting those 16 columns we get 50% to 100% increase in time while compared to the table with just those 16 columns.

I have uploaded a git diff fixing this issue. Kindly look into it.  
[cstore_large_col.diff.zip](https://github.com/citusdata/cstore_fdw/files/779100/cstore_large_col.diff.zip)

mtuncer:
A user reported performance issue if all stripe skip lists are
deserialized even if only a fraction of them is selected. 

Fixes #136

### DIFF
--- a/cstore_reader.c
+++ b/cstore_reader.c
@@ -54,6 +54,7 @@ static StripeSkipList * LoadStripeSkipList(FILE *tableFile,
 										   StripeMetadata *stripeMetadata,
 										   StripeFooter *stripeFooter,
 										   uint32 columnCount,
+										   bool *projectedColumnMask,
 										   Form_pg_attribute *attributeFormArray);
 static bool * SelectedBlockMask(StripeSkipList *stripeSkipList,
 								List *projectedColumnList, List *whereClauseList);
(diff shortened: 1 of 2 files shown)


cstore_reader.c
```diff
@@ -684,6 +699,13 @@ LoadStripeSkipList(FILE *tableFile, StripeMetadata *stripeMetadata,
 		ColumnBlockSkipNode *columnSkipList = NULL;
 		uint32 blockIndex = 0;
 
+		/* no need to create ColumnBlockSkipList if the column is not selected */
+		if (!projectedColumnMask[columnIndex] && columnIndex > 0)
```
pykello: Instead of doing `columnIndex > 0` here and later, can we just add `rowCount` to `struct StripeSkipList` and populate it at `LoadStripeSkipList()` and remove `StripSkipListRowCount()`? This way reader of code won't wonder why `columnIndex > 0` is needed.

mtuncer: We still need to make firstColumn check even if we do that.
replaced ```columnIndex == 0```  and ```columnIndex > 0``` checks with equivalent ```firstColumn``` expressions

mtuncer:
@pykello  ping

pykello:
Looks good. Merging.

Issue #124, opened 2016-09-29
Check if $PGDATA/cstore_fdw/<oid> folder exists


Dezzsoke:
Hello,

I want to have cstore_fdw to part of all the databases I create, so I installed this to template1 database.

When creating a new database it doesn't create a folder under cstore_fdw, and I receive an error:
```ERROR:  could not open file "/data1/pgsql/data/cstore_fdw/2813186/2813202" for writing: No such file or directory```

To fix this I created a folder with the oid name of my database, and this fixed the problem.

Can we enhance the fdw to check this?

Also what would be the command I have to execute to create that folder from the database (i.e. using SQL)

Regards,
  Balazs

mtuncer:
Hey @Dezzsoke 

cstore_fdw creates this folder when you create the server after creating extension. 

I was not able to reproduce this issue when I run it locally. Could you provide more context please ?

Regarding of folder creation, AFAIK postgresql does not have build-in support for that. However, you could use plpython to do that. Please see http://stackoverflow.com/questions/34877345/postgresql-trigger-with-python-to-create-folders-based-on-new-field-records

Dezzsoke:
To reproduce:

Log into the template1 database.
Install cstore_fdw there.
Create a server there.

Try to use cstore -> Success.

Create a new database. (it will be based on template1)
connect to that database.
Try to use cstore -> failure.
Create a server -> after this cstore works.

Dezzsoke:
I ended up writing a shell script that creates the <oid> folder after database creation, this solved the issue. (Not that ugly workaround)

mtuncer:
We are migrating to postgresql internal files as described in issue #138 .

Work is currently being reviewed. This issue will be resolved with it.

mtuncer:
@Dezzsoke I am adding a fix that would make sure that directory exists before attempting to create files. However, this still can not address following scenario
- cstore server is created in template1
- cstore table is created in template1
- some data is inserted into newly created cstore table in template1
- a new database db1 is created out of template1
- connected to new db1
- db1 seem to contain the foreign table, however it does not have anything in it.

I would rather prevent creation of cstore table inside a template database then try to fix this.

Dezzsoke:
@mtuncer I think preventing cstore table creation in template databases is acceptable.

mtuncer:
I am keeping this open due to template issue. cstore should prevent table creation when issued inside a template database.

[...]

Issue #129, opened 2016-12-08
Dropping a database does not remove the cstore files 


jan-schulz-k24:
We have a mirror script which copies a database from production to a mirror (dump+restore to a temp database) and also copies over the cstore files (on the production host, copy over the cstore file for the DB to a temp directory with schema+table based names, rsync to mirror and move to the new ID based names in the cstore_fdw folder) . We then move the temp database over to the mirror database using:

```sql
DROP DATABASE IF EXISTS mirror;
ALTER DATABASE temp_db RENAME TO mirror;"
```

Unfortunately, this doesn't remove the cstore files from the old mirror database, resulting in leaked cstore files for each mirror run :-/

Script to reproduce:
```sql
CREATE DATABASE whatever;

\c whatever;

CREATE EXTENSION IF NOT EXISTS cstore_fdw;
CREATE SERVER cstore_server FOREIGN DATA WRAPPER cstore_fdw;

CREATE SCHEMA foo;

CREATE FOREIGN TABLE foo.bar (
  i INT
)
SERVER cstore_server;

-- show the IDs for the cstore files in the current database
SELECT
  (SELECT oid
   FROM pg_database
   WHERE datname = current_database()) AS dboid,
  ftrelid,
  nspname,
  relname
FROM pg_foreign_table
  JOIN pg_class ON pg_class.oid = ftrelid
  JOIN pg_namespace ON pg_namespace.oid = relnamespace;

\c postgres;

DROP DATABASE whatever;
```

Running it via `cat script | psql postgres` will show this:

```
$ cat script | psql postgres
[...]
 dboid  | ftrelid | nspname | relname 
--------+---------+---------+---------
 854324 |  854337 | foo     | bar
(1 row)
DROP DATABASE
$ ls -la /var/lib/postgresql/9.5/main/cstore_fdw/*/* |grep 854324
-rw------- 1 postgres postgres         0 Dez  8 11:50 /var/lib/postgresql/9.5/main/cstore_fdw/854324/854337
-rw------- 1 postgres postgres        26 Dez  8 11:50 /var/lib/postgresql/9.5/main/cstore_fdw/854324/854337.footer
```

-> The database is droped, but the cstore files for that database are still there.

mtuncer:
Hey @jan-schulz-k24 

thanks for detailed reporting on the problem.  We will look into it.

PR #149, opened 2017-06-16
Directory creation and cleanup after database operations

mtuncer:
This PR addresses an issue when a foreign server is created inside a template database, newly created databases can not create foreign tables due to missing folder. This fix makes sure required directroy exists prior to creating files in it.

This also makes sure that cstore files are removed after a database or cstore extension is dropped out of current database.

There are 2 mishaps that we can't address
- if the cstore table is created with file name option, there is no way of removing it when a database/extension is dropped
- if the active database does not have cstore extension created, then we can't reclaim resources of the dropped database.

Fixes : #124 , #129

### DIFF
--- a/cstore_fdw.c
+++ b/cstore_fdw.c
@@ -28,6 +28,7 @@
 #include "catalog/pg_foreign_table.h"
 #include "catalog/pg_namespace.h"
 #include "commands/copy.h"
+#include "commands/dbcommands.h"
 #include "commands/defrem.h"
 #include "commands/event_trigger.h"
 #include "commands/explain.h"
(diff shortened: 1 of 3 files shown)


cstore_fdw.c
```diff
@@ -1003,6 +1068,51 @@ CreateDirectory(StringInfo directoryName)
...
+{
+	StringInfo cstoreDirectoryPath = makeStringInfo();
+	StringInfo cstoreDatabaseDirectoryPath = makeStringInfo();
+
+	appendStringInfo(cstoreDirectoryPath, "%s/%s", DataDir, CSTORE_FDW_NAME);
+
+	appendStringInfo(cstoreDatabaseDirectoryPath, "%s/%s/%u", DataDir,
+					 CSTORE_FDW_NAME, databaseOid);
+
+	if (DirectoryExists(cstoreDatabaseDirectoryPath))
```
pykello: Can we use `rmtree()` from `postgres/port.h` to remove the directory tree here?

mtuncer: yes we can

pykello:
Looks good to me, :shipit:

PR #150, opened 2017-06-16
Check if the user has truncate privilege on the table

### LINKED ISSUE #148: Revoke TRUNCATE permission from certain user
Hi, 

My purpose is to prevent commands like TRUNCATE from being issued on a cstore table by a non-privileged user.
As a superuser I specified REVOKE TRUNCATE, but other database users still able to truncate a table.

Steps to reproduce the issue:

- Initial data and permissions preparation
```
postgres=# CREATE DATABASE truncate_db;
CREATE DATABASE
postgres=# \c truncate_db
You are now connected to database "truncate_db" as user "postgres".
truncate_db=# CREATE EXTENSION cstore_fdw;
CREATE EXTENSION
truncate_db=# CREATE SERVER cstore_srv FOREIGN DATA WRAPPER cstore_fdw;
CREATE SERVER
truncate_db=# CREATE SCHEMA truncate_schema;
CREATE SCHEMA
truncate_db=# CREATE FOREIGN TABLE truncate_schema.truncate_tbl (id integer) SERVER cstore_srv OPTIONS(compression 'pglz');
CREATE FOREIGN TABLE
truncate_db=# INSERT INTO truncate_schema.truncate_tbl SELECT generate_series(1,100);
INSERT 0 100
truncate_db=# SELECT count(*) FROM truncate_schema.truncate_tbl;
 count
-------
   100
(1 row)

truncate_db=# CREATE USER truncate_user;
CREATE ROLE
truncate_db=# GRANT USAGE ON SCHEMA truncate_schema TO truncate_user;
GRANT
truncate_db=# GRANT SELECT ON TABLE truncate_schema.truncate_tbl TO truncate_user;
GRANT
truncate_db=# REVOKE TRUNCATE ON TABLE truncate_schema.truncate_tbl FROM truncate_user;
REVOKE
```
- Switch user and try to truncate a table
```
truncate_db=> select current_user;
 current_user
---------------
 truncate_user
(1 row)

truncate_db=> SELECT count(*) FROM truncate_schema.truncate_tbl;
 count
-------
   100
(1 row)

truncate_db=> truncate truncate_schema.truncate_tbl;
TRUNCATE TABLE
truncate_db=> SELECT count(*) FROM truncate_schema.truncate_tbl;
 count
-------
     0
(1 row)
```

Cheers.

mtuncer:
We rely on postgres to do access control checks for all functions. However, since truncate for cstore tables is handled in process utility,  not forwarded to postgres, and we did not check if a user has truncate right on the table, we let unprivileged user to perform truncate.

This fix introduces acl check prior to truncating files, and reports error if user has failed the check.

Fixes #148

### DIFF
--- a/cstore_fdw.c
+++ b/cstore_fdw.c
@@ -94,7 +94,8 @@ static uint64 CopyOutCStoreTable(CopyStmt* copyStatement, const char* queryStrin
 static void CStoreProcessAlterTableCommand(AlterTableStmt *alterStatement);
 static List * DroppedCStoreFilenameList(DropStmt *dropStatement);
 static List * FindCStoreTables(List *tableList);
-static void TruncateCStoreTables(List *cstoreTableList);
+static List * OpenRelationsForTruncate(List *cstoreTableList);
+static void TruncateCStoreTables(List *cstoreRelationList);
 static void DeleteCStoreTableFiles(char *filename);
 static void InitializeCStoreTableFile(Oid relationId, Relation relation);
 static bool CStoreTable(Oid relationId);
@@ -350,6 +351,9 @@ CStoreProcessUtility(Node * parseTree, const char *queryString,
 		List *allTablesList = truncateStatement->relations;
 		List *cstoreTablesList = FindCStoreTables(allTablesList);
 		List *otherTablesList = list_difference(allTablesList, cstoreTablesList);
+		List *cstoreRelationList = OpenRelationsForTruncate(cstoreTablesList);
+		ListCell *cstoreRelationCell = NULL;
+
 		if (otherTablesList != NIL)
 		{
 			truncateStatement->relations = otherTablesList;
@@ -358,7 +362,13 @@ CStoreProcessUtility(Node * parseTree, const char *queryString,
 								  destReceiver, completionTag);
 		}
 
-		TruncateCStoreTables(cstoreTablesList);
+		TruncateCStoreTables(cstoreRelationList);
+
+		foreach(cstoreRelationCell, cstoreRelationList)
+		{
+			Relation relation = (Relation) lfirst(cstoreRelationCell);
+			heap_close(relation, AccessExclusiveLock);
+		}
 	}
 	else if (nodeTag(parseTree) == T_AlterTableStmt)
 	{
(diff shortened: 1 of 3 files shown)


cstore_fdw.c
```diff
@@ -300,6 +301,9 @@ CStoreProcessUtility(Node * parseTree, const char *queryString,
 		List *allTablesList = truncateStatement->relations;
 		List *cstoreTablesList = FindCStoreTables(allTablesList);
 		List *otherTablesList = list_difference(allTablesList, cstoreTablesList);
+
+		CheckTruncatePermissions(cstoreTablesList);
```
pykello: If permissions of a table change by a concurrent session between call to CheckTruncatePermissions() and TruncateCStoreTables(), do we see it here?

(1) PostgreSQL's ExecuteTruncate() seems to do permission check & truncation while holding an AccessExclusiveLock, so I guess permissions cannot change after permission check and before end of TRUNCATE.
(2) Not caused by this change, but PostgreSQL grabs a lock on *all* relations before starting truncating the first table. In cstore_fdw we lock, truncate, release tables one by one.

So our behaviour here is not similar to PostgreSQL here.  I don't think this is a big issue, just trying to understand.

@anarazel what do you think? Can not having same semantics as PostgreSQL here cause much harm? (I'm sure these 2 aren't the only issues)

anarazel: > (1) PostgreSQL's ExecuteTruncate() seems to do permission check & truncation while holding an AccessExclusiveLock, so I guess permissions cannot change after permission check and before end of TRUNCATE.

I'm not particularly concerned about that race. But we really *should* take an AccessExclusiveLock during the check (or just acquire one before), because otherwise there's lock-upgrade issues.

>  (2) Not caused by this change, but PostgreSQL grabs a lock on all relations before starting truncating the first table. In cstore_fdw we lock, truncate, release tables one by one.

I don't think we "release" locks (that's what you mean, right) at all?

pykello: In TruncateCStoreTable() in cstore_fdw.c, doesn't heap_open() acquire a lock and heap_close() release that lock?

anarazel: > In TruncateCStoreTable() in cstore_fdw.c, doesn't heap_open() acquire a lock and heap_close() release that lock?

Indeed. And that's completely wrong.  I was only looking at CheckTruncatePermissions() which acquires a lock (of the too low level), and doesn't release it.

mtuncer:
Changed locking semantics
- renamed ACL checking function so that it checks access control, locks with AccessExclusiveLock, and returns relation list.
- Truncate functions now accepts list of previously locked relations, truncates, and unlocks them.

cstore_fdw.c
```diff
@@ -786,20 +789,63 @@ FindCStoreTables(List *tableList)
...
-static void
-TruncateCStoreTables(List *cstoreTableList)
+/* 
+ * OpenTruncatedRelations opens and locks relations for tables to be truncated.
+ *
+ * It also performs a permission checks to see if the user has truncate privilege
+ * on tables.
+ */
+static List *
+OpenTruncatedRelations(List *cstoreTableList)
```
pykello: naming suggestion:‌ OpenRelationsForTruncate()?

cstore_fdw.c
```diff
@@ -358,7 +361,7 @@ CStoreProcessUtility(Node * parseTree, const char *queryString,
 								  destReceiver, completionTag);
 		}
 
-		TruncateCStoreTables(cstoreTablesList);
+		TruncateCStoreTables(cstoreRelationList);
```
pykello: `heap_close()` the relations after truncation? (See `ExecuteTruncate()` in tablecmds.c)

mtuncer: ```TruncateCStoreTables()``` does heap_close. Do you mean closing relations outside ```TruncateCStoreTables()``` ?

cstore_fdw.c
mtuncer: sounds better

cstore_fdw.c
pykello: Sorry, I‌ had missed that. Maybe we should close outside `TruncateCStoreTables()` as you suggested so future readers of code don't miss that.

pykello:
Looks good to me. We can merge.

PR #152, opened 2017-07-17
Fix PG10 beta2 break on predicate_refuted_by function call

mtuncer:
PG10 Beta2 adds ```clause_is_check``` parameter to predicate_refuted_by function.

### DIFF
--- a/cstore_reader.c
+++ b/cstore_reader.c
@@ -789,7 +789,11 @@ SelectedBlockMask(StripeSkipList *stripeSkipList, List *projectedColumnList,
 							 blockSkipNode->maximumValue);
 
 			constraintList = list_make1(baseConstraint);
+#if (PG_VERSION_NUM >= 100000)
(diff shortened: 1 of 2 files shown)


cstore_reader.c
```diff
@@ -789,7 +789,11 @@ SelectedBlockMask(StripeSkipList *stripeSkipList, List *projectedColumnList,
 							 blockSkipNode->maximumValue);
 
 			constraintList = list_make1(baseConstraint);
+#if (PG_VERSION_NUM >= 100000)
```
pykello: Can we add PG10 to travis.yaml so this is automatically tested?

mtuncer: Last time I could not get it to work. It looks like there has been some changes in tools project. Trying again.

pykello:
Looks good. Merging.

acscott:
Cool!

Issue #132, opened 2016-12-14
Cstore crashing server when enabled, not used


dragomirr:
Hi Citus people,

We have very unusual case. Our postgres process crashes every time when we enable cstore_fdw. We don't use it, just build dependencies, put cstore_fdw in postgresl.conf file and create extension. 
Server is postgres 9.6 and cstore_fdw is 1.5 on CentOS 7.
We get  server crashes on every insert, even though cstore is just enabled and not used. We tested that over last few days and we are sure cstore is causing problems, cause in the moment we remove it from shared_preload_libraries and restart server, everything works as expected. We get segmentation faults and server crashes and reason is "current transaction and exit, because another server process exited abnormally and possibly corrupted shared memory" and our postgres log is full of these lines. After that database get in recovery mode. We also have a lot of 
`<p>LOG:  all server processes terminated; reinitializing</p>                           
<p>LOG:  database system was interrupted; last known up at 2016-12-13 12:47:07 UTC</p> 
<p>LOG:  database system was not properly shut down; automatic recovery in progress</p>
<p>LOG:  redo starts at 84/2DFFE8A8</p>                                                
` 
Our query is `<p>insert into posttag (domainid, postid, tagid)</p>
<p>select distinct t.domainid, t.postid, a.tagid</p>
<p>join tag a on t.domainid = a.domainid and t.tag = a.name</p>
<p>where 1=1</p>
<p>and t.domainid = p_domainid</p>
<p>and t.item = p_item</p>
<p>and t.processid = p_processID</p>
<p>and not exists (select 1 from posttag where domainid = t.domainid and postid = t.postid and tagid = a.tagid)</p>` and our conf is (writing part of conf that seems related)`
<p>shared_buffers = 4GB</p>
<p>work_mem = 1536MB</p>
<p>maintenance_work_mem = 1536MB</p>
<p>shared_preload_libraries = 'cstore_fdw'</p>
<p>effective_cache_size = 30GB</p>`

mtuncer:
Hello @dragomirr 

That is a weird case, insert query on a different table should not trigger cstore_fdw at al.

Could you get a call stack to see where it crashes ? Also, could you double check if cstore is build against correct postgresql version in case there are multiple PG instances.

mtuncer:
@dragomirr 
So far I am not able to reproduce it locally.
My setup
- centos-7.2.1511, kernel 3.10
- postgresql 9.6.1 installed from postresql yum repo
[...]

PR #182, opened 2018-03-28
fix a segfault when processing truncate in cached plans

orgrim:
Hello,

CStoreProcessUtility() removes its relations from the list of relations of
truncate statements before passing it to standard_ProcessUtility(). When
the truncate is run from a pl/pgsql function, its plan can be cached.
Since the filtered list of relations created by CStoreProcessUtility() is
freed after the execution of the truncate, it corrupts the cached plan.
To avoid this, just restore the list of relations of the node when done.

The testcase from #132 created by @emiel (https://gist.github.com/emiel/4d4fd8c5c5ba9173e3175a08bb055be4) reproduces the problem. The problem disappears with this fix, but I don't know if it is enough to close the whole issue.

Best regards,

### DIFF
--- a/cstore_fdw.c
+++ b/cstore_fdw.c
@@ -62,7 +62,7 @@
 #if PG_VERSION_NUM >= 100000
 #define CALL_PREVIOUS_UTILITY(parseTree, queryString, context, paramListInfo, \
 							  destReceiver, completionTag) \
-	PREVIOUS_UTILITY(plannedStatement, queryString, context, paramListInfo, \
+	PREVIOUS_UTILITY(localPlannedStatement, queryString, context, paramListInfo, \
 					 queryEnvironment, destReceiver, completionTag)
 #else
 #define CALL_PREVIOUS_UTILITY(parseTree, queryString, context, paramListInfo, \
@@ -276,7 +276,10 @@ CStoreProcessUtility(Node * parseTree, const char *queryString,
 #endif
 {
 #if PG_VERSION_NUM >= 100000
-	Node *parseTree = plannedStatement->utilityStmt;
+	/* truncate processing needs to modify the parseTree of the
+	 * planned Statement, therefore we need to work on a copy */
+	PlannedStmt *localPlannedStatement = copyObjectImpl(plannedStatement);
+	Node *parseTree = localPlannedStatement->utilityStmt;
 #endif
 
 	if (nodeTag(parseTree) == T_CopyStmt)
@@ -347,7 +350,12 @@ CStoreProcessUtility(Node * parseTree, const char *queryString,
 	}
 	else if (nodeTag(parseTree) == T_TruncateStmt)
 	{
-		TruncateStmt *truncateStatement = (TruncateStmt *) parseTree;
+#if PG_VERSION_NUM >= 100000
+		TruncateStmt *truncateStatement =  (TruncateStmt *) parseTree;
+#else
+		/* we are modifying the parseTree, therefore we will work on a copy */
+		TruncateStmt *truncateStatement =  (TruncateStmt *) copyObject(parseTree);
+#endif
 		List *allTablesList = truncateStatement->relations;
 		List *cstoreTablesList = FindCStoreTables(allTablesList);
 		List *otherTablesList = list_difference(allTablesList, cstoreTablesList);
@@ -358,7 +366,7 @@ CStoreProcessUtility(Node * parseTree, const char *queryString,
 		{
 			truncateStatement->relations = otherTablesList;
 
-			CALL_PREVIOUS_UTILITY(parseTree, queryString, context, paramListInfo,
+			CALL_PREVIOUS_UTILITY((Node *) truncateStatement, queryString, context, paramListInfo,
 								  destReceiver, completionTag);
                         /* restore the former relation list. Our
                          * replacement could be freed but still needed
--- a/cstore_fdw.c
+++ b/cstore_fdw.c
@@ -368,11 +368,6 @@ CStoreProcessUtility(Node * parseTree, const char *queryString,
 
 			CALL_PREVIOUS_UTILITY((Node *) truncateStatement, queryString, context, paramListInfo,
 								  destReceiver, completionTag);
-                        /* restore the former relation list. Our
-                         * replacement could be freed but still needed
-                         * in a cached plan. A truncate can be cached
-                         * if run from a pl/pgSQL function */
-                        truncateStatement->relations = allTablesList;
 		}
 
 		TruncateCStoreTables(cstoreRelationList);
--- a/cstore_fdw.c
+++ b/cstore_fdw.c
@@ -360,6 +360,11 @@ CStoreProcessUtility(Node * parseTree, const char *queryString,
 
 			CALL_PREVIOUS_UTILITY(parseTree, queryString, context, paramListInfo,
 								  destReceiver, completionTag);
+                        /* restore the former relation list. Our
+                         * replacement could be freed but still needed
+                         * in a cached plan. A truncate can be cached
+                         * if run from a pl/pgSQL function */
+                        truncateStatement->relations = allTablesList;
 		}
 
 		TruncateCStoreTables(cstoreRelationList);
--- a/expected/truncate.out
+++ b/expected/truncate.out
@@ -155,6 +155,27 @@ SELECT * from cstore_truncate_test;
 ---+---
 (0 rows)
 
+-- test if a cached truncate from a pl/pgsql function works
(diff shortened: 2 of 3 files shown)


mtuncer:
Thanks for the PR.

1. Could you provide a test case that reproduces this issue so that we can verify that this fix solves it ?
2. Could you target it to `develop` branch instead of `master` branch ? We are merging it to master only when releasing a new version.
3. You have restored original table list only if there is a regular table in the truncate list. Table list is not restored if it only consist of cstore tables. Is this intentional ?

thanks again.

kostiantyn-nemchenko:
Hi,

Just had this bug happen to me as well. Here are results of the test case by @emiel that reproduces mentioned issue on my side.
<details>
 <summary>Testcase results</summary>

```sql
postgres=# CREATE DATABASE cstore_issue;
CREATE DATABASE
postgres=# \c cstore_issue
You are now connected to database "cstore_issue" as user "postgres".
cstore_issue=# CREATE EXTENSION cstore_fdw VERSION "1.6" ;
CREATE EXTENSION
cstore_issue=# create table queue
cstore_issue-# (
cstore_issue(#     id integer,
cstore_issue(#     diff integer
cstore_issue(# );
CREATE TABLE
cstore_issue=# create table cache
cstore_issue-# (
cstore_issue(#     id integer primary key,
cstore_issue(#     count integer not null,
cstore_issue(#     last_update_at timestamptz
cstore_issue(# );
CREATE TABLE
cstore_issue=# create or replace function process_queue()
cstore_issue-# returns bool language plpgsql
cstore_issue-# as $body$
cstore_issue$# begin
cstore_issue$#
cstore_issue$#     insert into cache (id, count, last_update_at)
cstore_issue$#         select id, sum(diff), now()
cstore_issue$#         from queue
cstore_issue$#         group by id
cstore_issue$#     on conflict (id) do update set
cstore_issue$#         count = cache.count + EXCLUDED.count,
cstore_issue$#         last_update_at = EXCLUDED.last_update_at
cstore_issue$#         ;
cstore_issue$#
cstore_issue$#      -- removing the truncate line resolves the issue
cstore_issue$#      truncate queue;
cstore_issue$#
cstore_issue$#     return true;
cstore_issue$# end;
cstore_issue$# $body$;
CREATE FUNCTION
cstore_issue=# select process_queue(); select process_queue(); select process_queue();
 process_queue
---------------
 t
(1 row)

server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
```
</details>
<br>
Can confirm that the segmentation fault doesn't occur anymore after applying the suggested fix by @orgrim

orgrim:
Hello,

1. A simplified testcase, that works on all supported versions of PostgreSQL:

```sql
CREATE TABLE IF NOT EXISTS public.testcase (
	i int
);

CREATE OR REPLACE FUNCTION public.testcase_truncate()
RETURNS void
AS $$
BEGIN
	INSERT INTO public.testcase (i) VALUES (1);
	TRUNCATE public.testcase;
END;$$
LANGUAGE plpgsql;

SELECT public.testcase_truncate(); 
-- at this point the plan of the truncate in the function should be cached
-- and the next call should trigger the segfault
SELECT public.testcase_truncate();
-- repeat until it crashes, but it should die on the second call
SELECT public.testcase_truncate();
SELECT public.testcase_truncate();
```

2. Concerning the branches, I see a `develop` and a `develop_v1x`, I guess `develop_v1x` will become the release branch for v1 when v2 is out?

I can rebase on `develop` and update the PR, but would there be à 1.6.1 version soon? or before v2?

3. the list of relation is modified only if there are non-cstore relations just before passing it to the previous utility hook function, so we only have to restore when the other utilities have done their work:

```c
if (otherTablesList != NIL)
{
        truncateStatement->relations = otherTablesList;

        CALL_PREVIOUS_UTILITY(parseTree, queryString, context, paramListInfo,
                                                                  destReceiver, completionTag);
                        /* restore the former relation list. Our
                         * replacement could be freed but still needed
                         * in a cached plan. A truncate can be cached
                      * if run from a pl/pgSQL function */
        truncateStatement->relations = allTablesList;
}
```

mtuncer:
After discussion with @marcocitus  we had a different fix. But I do not want to hijack this pr.

Could you change the line
```
TruncateStmt *truncateStatement = (TruncateStmt *) parseTree;
```
to
```
/* we are modifying the parseTree, therefore we will work on a copy */
TruncateStmt *truncateStatement = (TruncateStmt *) copyObject(parseTree);
```

Also if you could add a test case to truncate.sql.

Since this is important enough, we will have a patch release containing this.

Thanks

mtuncer:
@orgrim  clarification: don't bother rebasing to develop, this will directly go into master.

orgrim:
Hello,

I've just added the regressions test.

Making a with copyObject() would feel better than modifying/restoring the variable, but it would make the patch more complicated. We would have to copy the plannedStatement object for PG >= 10, this adds more preprocess tests.

orgrim:
@mtuncer please review the last commit that uses a copy instead of restoring the input variable

Thanks

cstore_fdw.c
```diff
@@ -62,7 +62,7 @@
 #if PG_VERSION_NUM >= 100000
 #define CALL_PREVIOUS_UTILITY(parseTree, queryString, context, paramListInfo, \
 							  destReceiver, completionTag) \
-	PREVIOUS_UTILITY(plannedStatement, queryString, context, paramListInfo, \
+	PREVIOUS_UTILITY(localPlannedStatement, queryString, context, paramListInfo, \
```
mtuncer: we should not need this

cstore_fdw.c
```diff
@@ -347,7 +350,12 @@ CStoreProcessUtility(Node * parseTree, const char *queryString,
 	}
 	else if (nodeTag(parseTree) == T_TruncateStmt)
 	{
-		TruncateStmt *truncateStatement = (TruncateStmt *) parseTree;
+#if PG_VERSION_NUM >= 100000
+		TruncateStmt *truncateStatement =  (TruncateStmt *) parseTree;
+#else
+		/* we are modifying the parseTree, therefore we will work on a copy */
+		TruncateStmt *truncateStatement =  (TruncateStmt *) copyObject(parseTree);
```
mtuncer: I think we are good for just copying parseTree here, there should not be any need for making a copy of plannedStatement.

cstore_fdw.c
```diff
@@ -358,8 +366,13 @@ CStoreProcessUtility(Node * parseTree, const char *queryString,
...
 			truncateStatement->relations = otherTablesList;
 
-			CALL_PREVIOUS_UTILITY(parseTree, queryString, context, paramListInfo,
+			CALL_PREVIOUS_UTILITY((Node *) truncateStatement, queryString, context, paramListInfo,
 								  destReceiver, completionTag);
+                        /* restore the former relation list. Our
+                         * replacement could be freed but still needed
+                         * in a cached plan. A truncate can be cached
+                         * if run from a pl/pgSQL function */
+                        truncateStatement->relations = allTablesList;
```
mtuncer: this is not needed when you make a copy of parse tree.

cstore_fdw.c
orgrim: It is needed since this variable is not passed by the macro call. We have to point it to the modified copy of the PlannedStmt so that other ProcessUtility hooks do not see cstore tables.

cstore_fdw.c
orgrim: If we just copy the parseTree for PG >= 10, we have to modify the input PlannedStmt and restore it after the call to CALL_PREVIOUS_UTILITY.

cstore_fdw.c
orgrim: Indeed. Fixed in a13ace227f139622724f347b33c063a3d1aa19e5

orgrim:
@mtuncer, I've just double checked. We have to copy the whole variable that is passed to the previous hook. For PG < 10, passing the copied parseTree works. For PG >= 10, since the variable name is hardcoded in CALL_PREVIOUS_UTILITY, wa have to modify CALL_PREVIOUS_UTILITY to use the copy and copy the planned statement.

cstore_fdw.c
mtuncer: I did not like this approach. We will end up performing copy when any utility call is made which could be a lot.

If we can't get away with only copying parseTree, then we should revert back to your original solution of restoring table list back, and ditch all the rest.

orgrim:
@mtuncer please review PR #184 that revert to copy of the input to CALL_PREVIOUS_UTILITY

Regards,

mtuncer:
just merged #184, closing this in favor of that one.

Thanks a lot for your efforts on investigating and fixing the issue.

PR #184, opened 2018-04-05
fix a segfault when processing truncate in cached plans (final version)

orgrim:
This is the final patch for the fix worked on in #182 

It is a rebase on master keeping the first 2 commits (fix and regression test) melted in one.


orgrim:
@mtuncer Thank you. Do have an ETA for a release 1.6.1?

mtuncer:
there is another patch in develop_1x that needs to be merged here. Than 1.6.1 should be good to go.
ETA is next few days, I need to spare some time to this.

Issue #161, opened 2017-08-31
Support for Parallel Scans


juliano777:
Columnar Storage + Parallel Scans = Conquers the Universe :)

masayuki038:
Did anyone try it? I want to know how to scan columnar data in parallel.

masayuki038:
I had a little challenge and succeeded.(on pg11beta2)
https://gist.github.com/masayuki038/6820f551f1a6d4e1ec7b3929ffad35ad

But I did not implement some methods to handle shared memory.(ex. EstimateDSMForeignScan)
Do anyone tell me how to estimate the shared memory size required for parallel scan?

juliano777:
Amazing!

PR #198, opened 2018-12-03
Mark foreign scan as parallel safe

### LINKED ISSUE #178: How to perform parallel execution on the `cstore_fdw` partition table ?
I was use `cstore_fdw`, and I created 20 partition table ,when i query the partition table ,I realize the query plan did not show the parallel.But when use the normal table ,the parallel is worked.

This make me confuse. Is there anyone can help?

marcocitus:
Fixes #178 

There's no reason why parallel scans of a cstore_fdw table would not be safe right now. Multiple processes can read from tables in parallel and there is no state kept between individual statements, which means that another process (like a parallel worker) can read everything that was written. 

Partially addresses #161 

Based it against master because develop does not have PG11 support


mtuncer:
This change should be committed to both `develop_v1x` and `develop` branches

PR #222, opened 2020-02-17
pg12 compatibility related changes

mtuncer:
- fix the build
- revisit data file naming method. (use the old logic in pre-pg12)

### DIFF
--- a/cstore_fdw.c
+++ b/cstore_fdw.c
@@ -43,7 +43,14 @@
 #include "optimizer/pathnode.h"
 #include "optimizer/planmain.h"
 #include "optimizer/restrictinfo.h"
+#if PG_VERSION_NUM >= 120000
+#include "access/heapam.h"
+#include "access/tableam.h"
+#include "executor/tuptable.h"
+#include "optimizer/optimizer.h"
+#else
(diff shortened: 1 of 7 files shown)


cstore_fdw.c
```diff
@@ -1553,22 +1553,15 @@ ValidateForeignTableOptions(char *filename, char *compressionTypeString,
...
 {
-	Relation relation = relation_open(foreignTableId, AccessShareLock);
-	RelFileNode relationFileNode = relation->rd_node;
-
-	Oid databaseOid = relationFileNode.dbNode;
-	Oid relationFileOid = relationFileNode.relNode;
-
+	Oid databaseOid = MyDatabaseId;
 	StringInfo cstoreFilePath = makeStringInfo();
 	appendStringInfo(cstoreFilePath, "%s/%s/%u/%u", DataDir, CSTORE_FDW_NAME,
```
serprex: Will there need to be upgrade instructions for this?

mtuncer: behavior is the same as long as you stay in the same PG version. We never officially supported upgrading PG.  So we should be fine there

